PR #1618, opened 2022-11-28
Move to cuprite for local system tests

danhalson:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

* Moves from selenium to cuprite, cuprite is faster and easier to configure and allows removing a lot of the complexity from our local test setup.

* N/A


danhalson:
This may be useful for future use and is only 3 test failures off working, but it makes no sense to create a new inconsistency between tc and tcc prior to handover, so am leaving.

PR #1642, opened 2023-01-16
font update

nyashamaphosa:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2247
* 
## What's changed?

*font change for h3 headings*


app/webpacker/stylesheets/components/_external-content.scss
```diff
@@ -13,6 +13,7 @@
   h2,
   h3 {
     @extend .heading-generic;
+    font-family: 'Objectivity', sans-serif;
```
danhalson, line 16: as discussed, I think this would be better set on `.govuk-heading-m` since I don't think we want this font on every h3, one to double check with @maddyjoe if you aren't sure, also might be worth doing a tiny design check to verify it's only impacted the expected headings?

danhalson (changes requested):
Please check but probably better to only update `govuk-heading-m`

app/webpacker/stylesheets/components/_external-content.scss
danhalson: I don't think the standard h3 should be using this font too?

danhalson (changes requested):

app/webpacker/stylesheets/components/_external-content.scss
nyashamaphosa: resolved too thank you 😊

danhalson (approved):
LGTM

PR #1643, opened 2023-01-16
escapes quotes

nyashamaphosa:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2211

## What's changed?

Escapes comment symbols in rating of a comment


app/controllers/concerns/curriculum/rateable.rb
```diff
@@ -35,7 +35,7 @@ def comment
       response = client.update_rating(
         id: request[:rating_id],
         key: :comment,
-        value: request[:comment],
+        value: request[:comment].gsub('"', '&quote;'),
```
danhalson, line 38: it would be more flexible to use something like `CGI.escapeHTML` (https://ruby-doc.org/stdlib-3.1.0/libdoc/cgi/rdoc/CGI/Escape.html)...

however maybe a question for @maddyjoe is whether this is ok? since there'll now be entities in the comments which will make them harder to read, perhaps we'd be better stripping quotes in a way that doesn't add alter the string

danhalson: could you also add some tests around this please

spec/support/shared_examples/curriculum/rateable_examples.rb
```diff
@@ -55,6 +55,12 @@
       post send(comment_path, rating_id: rating_id, comment: 'This is a test')
       expect(response).to have_http_status(:ok)
     end
+
+    it 'allows comments with double quotes to be escaped and submitted' do
+      stub_a_valid_request({ data: { update_rating: {} } }.to_json)
+      post send(comment_path, rating_id: rating_id, comment: '"<body style="background-color:powderblue;">')
+      expect(response).to have_http_status(:ok)
```
danhalson, line 62: might be just a little more robust to also do: `expect(Rating.last.comment).to eq('\"<body style=\"background-color:powederblue;\">')` since I'm not 100% confident about how well the api is handling errors at the moment, and I worry that you'll get an :ok status regardless

danhalson (changes requested):

spec/support/shared_examples/curriculum/rateable_examples.rb
nyashamaphosa: resolved thank you

nyashamaphosa: > might be just a little more robust to also do: `expect(Rating.last.comment).to eq('\"<body style=\"background-color:powederblue;\">')` since I'm not 100% confident about how well the api is handling errors at the moment, and I worry that you'll get an :ok status regardless

resolved thank you 😊

danhalson: sorry my suggested code had a typo in it `background-color:powederblue` should be `background-color:powderblue`, if you change that the test might pass?

nyashamaphosa: I am not sure the class Rating is exists in TC as it only captures the comment but not store it. However the expected change registers in the curriculum database.

danhalson: ah yes! I was forgetting this was TC

danhalson (approved):
LGTM

PR #1649, opened 2023-01-25
Add ERD Gem and auto-create ERD PDF on migrate

grega:
## Status

* Related to https://github.com/NCCE/teachcomputing.org-issues/issues/2167

## What's changed?

Adds the rails-erd Gem, creates a task to auto-generate `erd.pdf` whenever new migrations are applied, and add documentation.


danhalson (approved):
LGTM

PR #1652, opened 2023-02-01
Update the default dev/staging endpoint for Achiever Smart Connector

dcorking:
also update the stubs for spec/services/achiever/request_spec.rb for consistency

## Status

## Review progress:

~- [ ] Browser tested~
~- [ ] Front-end review completed~
- [x] Tech review completed

## Steps to perform after deploying ~to production~ _locally_

If you don't already have this value in `.env` then restart your docker containers with `yarn run stop` and `yarn run start` to pick up the new value.


danhalson (approved):
Looks good, guess this had been obfuscated locally as we were using the templates

PR #1653, opened 2023-02-01
Fix access to Achiever Smart Connector from a docker container without a proxy

dcorking:
## Status

## Review progress:

~- [ ] Browser tested~
~- [ ] Front-end review completed~
- [x] Tested by RPF developer
- [x] Tech review completed

## What's changed?

Storm doesn't have a proxy tunnel, as we share a fixed IPv4 address, so I removed PROXY_URL from the compose file, to allow this setting in the local `.env` to override it

I moved this value to `.env.defaults` so that the proxy still works for the RPF devs that need it.

I tested by running an achiever rake task in the web container's console:
```
root@9ae610137ec9 /app - (fix/default-smart-connector-host) $ bin/rails achiever:get_course_info
false
false
false
false
false
Achiever currently has 170 courses with 1 F2F & 0 online occurrences
Time: 3.836331785 s.
root@9ae610137ec9 /app - (fix/default-smart-connector-host) $
```

### DIFF
--- a/app/services/achiever/connection.rb
+++ b/app/services/achiever/connection.rb
@@ -5,7 +5,7 @@ class Achiever::Connection
     Faraday.new(url: ENV.fetch('ACHIEVER_V2_ENDPOINT')) do |conn|
       conn.adapter :net_http
       conn.request(:basic_auth, ENV.fetch('ACHIEVER_V2_USERNAME'), ENV.fetch('ACHIEVER_V2_PASSWORD'))
-      conn.proxy = ENV.fetch('PROXY_URL')
+      conn.proxy = ENV.fetch('PROXY_URL') # set PROXY_URL='' if you don't need a proxy
     end
   end
 end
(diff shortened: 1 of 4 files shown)


dcorking:
@danhalson Please test that this works for you, as the last think I'd want to do is break your team's development setups, and I didn't install my own proxy to try it.

app/services/achiever/connection.rb
```diff
@@ -5,7 +5,7 @@ def self.api
     Faraday.new(url: ENV.fetch('ACHIEVER_V2_ENDPOINT')) do |conn|
       conn.adapter :net_http
       conn.request(:basic_auth, ENV.fetch('ACHIEVER_V2_USERNAME'), ENV.fetch('ACHIEVER_V2_PASSWORD'))
-      conn.proxy = ENV.fetch('PROXY_URL')
+      conn.proxy = ENV.fetch('PROXY_URL') # set PROXY_URL='' if you don't need a proxy
```
danhalson, line 8: unless I'm misunderstanding the intention, you could just add a fallback to an empty string `ENV.fetch('PROXY_URL', '')` that way it doesn't need to be defined

PR #1660, opened 2023-02-16
Allow Sentry to be passed an environment ENV var

grega:
Environments can then optionally set `SENTRY_CURRENT_ENV` to a value (eg. `review`, `staging`, `qa`, `production`) in order to more clearly delineate where error reports are coming from.


danhalson (approved):
LGTM

PR #1675, opened 2023-03-04
Seed pathways with new moodle courses (replacing FL)

dcorking:
## Status

* Related to https://github.com/NCCE/teachcomputing.org-issues/issues/2299

## Review progress:

- [x] Tech review completed

## What's changed?

Add the relevant new Stem Learning online courses (Moodle) to the pathways.
1. CSA certificate: 5 pathways
2. Primary certificate: 2 pathways

One-off script to remove all the FutureLearn pathway activities (and a script to restore them just in case, made by copying the old seeds.)

Also removed the FutureLearn course from the pathway activity seeds. (FL pathway activities won't be restored by the seeds, and won't be in any newly seeded databases)

although new seeds are automatic you need to run a rake task to remove the old courses from the pathways and reseed them

    bin/rails fix_pathway_activities


db/seeds/pathways/cs_accelerator.rb
```diff
@@ -20,10 +20,10 @@
 activity = Activity.find_by(stem_course_template_no: 'e1e54959-db12-eb11-a813-000d3a86d545')
 pathway.pathway_activities.find_or_create_by(activity_id: activity.id) if activity
 
-activity = Activity.find_by(future_learn_course_uuid: '645ec51f-0b46-4102-a364-90647057f4f2')
+activity = Activity.find_or_create_by(stem_course_template_no: '83e7b049-a5b6-ed11-b597-0022481b59ce')
```
danhalson, line 23: was this meant to be a `find_or_create_by`? pathways should only be created once the activity seeds have been run and activities exist, so it seems better to do nothing here, or better would be to error perhaps? though I appreciate that's probably out of scope now

lib/tasks/rm_future_learn_from_pathways.rake
```diff
@@ -0,0 +1,105 @@
+namespace :rm_future_learn_from_pathways do
+  desc 'one-off task to remove the future learn courses from their pathways'
+  task remove: :environment do
+    puts 'Destroying pathway join records for all future learn courses'
```
danhalson, line 4: Happy to stand corrected if I'm missing something, but because pathways are predominantly a meta thing, we found it safest to just remove them all and run the seeds manually, like this:

https://github.com/NCCE/teachcomputing.org/blob/main/lib/tasks/fix_pathway_activities.rake#L6

dcorking: Thank you!

As a habit, I don't normally destroy everything and I didn't look at your 'fix' task as I didn't have anything to fix (yet) 😂 . That rake task looks like it will do what I need, so I'll delete the new one.

That means our recovery plan will require a rollback and running that task again.

danhalson (approved):
LGTM

PR #1697, opened 2023-03-23
Update titles of CP008 CP456 and CP005

dcorking:
closes https://github.com/NCCE/teachcomputing.org-issues/issues/2332

@maddyjoe would you like Storm or RPF to deploy this? And are you OK with someone changing these titles now (rather than waiting until NCCE2)?


danhalson (dismissed):
LGTM

PR #1702, opened 2023-04-05
Some cleanups incl RPF domain

dcorking:
## Status

* Related to https://github.com/NCCE/teachcomputing.org-issues/issues/2325

## Review progress:
- [x] Tech review completed

## What's changed?

1. Update DEFAULT_ADMIN_EMAIL for developers      
2. In specs: replace web@raspberrypi.org with web@teachcomputing.org            
3. Remove RPF references from BYPASS_OAUTH feature and from comments            
4. Rm unused assignments in certificate generator                               
5. Report absence of occurrences to Sentry                                      

## Steps to perform when deploying to production

- [x] create a web@teachcomputing.org user
```ruby
user = User.create(first_name: 'STEM Learning',last_name: 'Digital',email: 'web@teachcomputing.org', stem_achiever_contact_no: Random.uuid, stem_credentials_access_token: '0000', stem_credentials_refresh_token: '0000', stem_credentials_expires_at: DateTime.now, stem_user_id: '0000000')
```
- [x] deploy terraform PR https://github.com/NCCE/terraform/pull/182  to change the environment variable for the DEFAULT_ADMIN_EMAIL
- [ ] deploy this branch


dcorking:
- [ ] investigate ability to edit user in adminsitrate before prod deploy https://stem-learning.sentry.io/issues/4064402564/?referrer=slack

app/controllers/courses_controller.rb
```diff
@@ -71,6 +71,7 @@ def assign_start_date
         @start_date = start_date(@occurrences)
         @started = (@start_date <= Date.today)
       else # This shouldn't happen but some test data has no occurrences
+        Sentry.capture_message('Attempted to determine start date of a course with no occurrences', extra: { course_template_no:  @course&.course_template_no })
```
markjs, line 74: From further up in the file I don't think this `&.` is achieving anything since a `nil` value of `@course` would error elsewhere. May be an idea to remove it so it doesn't inadvertently hide a future regression.

config/initializers/content_security_policy.rb
```diff
@@ -7,8 +7,8 @@
 Rails.application.config.content_security_policy do |policy|
 #   policy.default_src :self, :https, 'youtube.com', 'www.youtube.com'
 #   policy.font_src    :self, :https, :data, 'fonts.gstatic.com'
-#   policy.img_src     :self, :https, :data, 'view.vzaar.com', 'blog.teachcomputing.org', 'futurelearn-assets.raspberrypi.org'
-#   policy.media_src   :self, :https, :data, 'view.vzaar.com', 'download.vzaar.com', 'futurelearn-assets.raspberrypi.org'
+#   policy.img_src     :self, :https, :data, 'view.vzaar.com', 'blog.teachcomputing.org'
+#   policy.media_src   :self, :https, :data, 'view.vzaar.com', 'download.vzaar.com'
```
markjs, line 11: What is the value in these commented out lines remaining in the project? There isn't a comment explaining why they're here or how to use them - perhaps that is worth adding if they're valuable?

markjs (commented):
Couple of small comments but otherwise looks good to me 👍

config/initializers/content_security_policy.rb
dcorking: It was temporarily disabled 2 years ago. https://github.com/NCCE/teachcomputing.org/pull/1160
I'll add a comment.

app/controllers/courses_controller.rb
markjs: Sounds reasonable. My concern was that the Sentry message would end up with a misleading `nil` value but you're right that the Sentry message erroring or not sending would be a more unhelpful regression. Happy for this to be left as is.

markjs (approved):

PR #1719, opened 2023-05-18
Suggested refactor of question bank resource partial

dcorking:
@nyashamaphosa when you pass a collection to render, it creates a local variable with the same name as the partial

Merge or adapt this code if you find it useful


dcorking:
the parent branch now implements this

PR #1725, opened 2023-06-01
Tighten ratings cookie flag settings

dcorking:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2373

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

The browser will be more strict with these flags set.

The secure flag is false in development and test where we don't use
https.

In development, if you are working with ratings, add this to your `.env`
```
SECURE_COOKIES=false
```

### DIFF
--- a/.env.defaults
+++ b/.env.defaults
@@ -65,3 +65,6 @@ DEFAULT_ADMIN_EMAIL=web@teachcomputing.org
 
 # This must exist in .env for docker-compose to detect it
 ENV_FILE=.env
+
+# set to off to turn off secure cookies; not needed in production
+SECURE_COOKIES=off
--- a/.env.test
+++ b/.env.test
@@ -35,3 +35,6 @@ BYPASS_OAUTH=false
 BYPASS_ADMINISTRATE_CF_AUTH=false
 
 DEFAULT_ADMIN_EMAIL=web@teachcomputing.org
+
+# set to off to turn off secure cookies
+SECURE_COOKIES=off
--- a/app/controllers/concerns/curriculum/rateable.rb
+++ b/app/controllers/concerns/curriculum/rateable.rb
@@ -86,7 +86,7 @@ module Curriculum
       raw_cookie = cookies.encrypted[:ratings]
       ratings = raw_cookie.nil? ? [] : JSON.parse(raw_cookie)
       ratings << id
-      cookies.encrypted[:ratings] = { value: JSON.generate(ratings), expires: 1.month }
+      cookies.encrypted[:ratings] = { value: JSON.generate(ratings), expires: 1.month, httponly: true, secure: Rails.configuration.secure_cookies }
     end
   end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -27,5 +27,8 @@ module TeachComputing
     config.i18n.load_path += Dir[Rails.root.join('config', 'locales', '**/*.{rb,yml}').to_s]
 
     config.mylearning_dashboard_url = ENV['MYLEARNING_DASHBOARD_URL']
+
+    # default is true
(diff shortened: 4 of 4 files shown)


dcorking:
(Add @caleb-t-owens as a reviewer )

Caleb-T-Owens:
in https://github.com/NCCE/teachcomputing.org/blob/main/config/environments/production.rb#L133 and similarly in the staging config file. Would it make sense to move the session store config into the application.rb file so that `secure_cookies` flag can be observed in all environments for the session cookie? The staging and production environments would maintain the same behaviour, but it would allow SSL-enabled development environments to have a secure session cookie.

dcorking:
@Caleb-T-Owens yes. Easy to do. I think it is out of scope of this ticket. So, even simpler, I recommend that when we make an SSL-enabled development environment, we undo this commit, and set `config.force_ssl = true` in `application.rb` which [sets the secure flag to all cookies by default](https://api.rubyonrails.org/v6.1/classes/ActionDispatch/SSL.html)

Caleb-T-Owens:
For QA I have visited the two page types which contain review buttons. On these pages, I have clicked on a thumb direction and observed that the network response that by browser gets is indeed setting a secure cookie.

I have observed two unexpected behaviours of this feature, but when compared to what is in production the behaviour is in line, so this shouldn't prevent the deployment of this change, but I felt it worthwhile mentioning here just so someone who is more familiar with the history of this feature could confirm whether or not it was expected behaviour and if it needs fixed or not.

Firstly:
https://github.com/NCCE/teachcomputing.org/assets/13354155/f2ece4fa-b306-46ed-baac-6e0a40671b20

When I visit a page with ratings on it, when I then click on the thumbs up, it correctly gives me more options for feedback. When I then reload the page, I no longer can provide further options for feedback.

I would expect this to continue providing the option for further feedback when I reload the page.


Secondly:
https://github.com/NCCE/teachcomputing.org/assets/13354155/02a2f173-a0b1-41f5-a7b6-63d6d31b5a22

When I visit a page with ratings on it, when I click on the thumbs up, when I navigate to another page and then navigate back using the browser back button, I am presented with an option to click the thumbs buttons again which don't allow me to click them again.

I would expect the page to move to the state where the buttons are no longer available or they are asking for further feedback

Caleb-T-Owens:
> @Caleb-T-Owens yes. Easy to do. I think it is out of scope of this ticket. So, even simpler, I recommend that when we make an SSL-enabled development environment, we undo this commit, and set `config.force_ssl = true` in `application.rb` which [sets the secure flag to all cookies by default](https://api.rubyonrails.org/v6.1/classes/ActionDispatch/SSL.html)

I'm surprised we don't force_ssl already. Is there a business reason as to why it was set up without SSL enforced?

If there is no business reason, having force_ssl enabled would be ideal, not just for the security of cookies but the app in general. It might make sense to hinge it off a DISABLE_SSL env variable in case someone was struggling to get SSL to work in development though

dcorking:
@Caleb-T-Owens I have not heard any business reason for not forcing SSL, and I can't think of any technical reason in the infrastructure. I think you make a persuasive case to add this to the backlog: the main cost is that 2 or 3 devs will update their dev environments to suit.

dcorking:
@dapo-olaitan I think Caleb's discovery about [the clunky behaviour of the ratings widget (above)](https://github.com/NCCE/teachcomputing.org/pull/1725#issuecomment-1602176451) is something that might be worth you making a ticket for in a future sprint: it isn't ideal. Unfortunately I don't know who the business stakeholder for the ratings feature is, though it does feel like a significant rewrite might be needed to eliminate these behaviours.

Caleb-T-Owens (approved):

PR #1736, opened 2023-06-27
Fix source metadata from background job in pending certificate

dcorking:
## Status

* Related to 
  - https://github.com/NCCE/teachcomputing.org-issues/issues/2398
  - https://github.com/NCCE/teachcomputing.org-issues/issues/2334

## Review progress:
- [x] Tech review completed

## What's changed?

- Fix misleading source metadata from background job in pending certificate to improve debugging

(this data is stored in Postgres as JSON in `UserProgrammeEnrolmentTransition#metadata` )


Caleb-T-Owens (approved):
I'm happy with this change code-wise and don't think it warrants any QA of the app as this is only changing information that gets moved into a logger

PR #1747, opened 2023-07-20
2387/mark teachers as qualified

Caleb-T-Owens:
## Status

* Current status Ready for (front-end / tech) review
* Related to #2387

## Review progress:

- [N/A] Browser tested 
- [N/A] Front-end review completed
- [ ] Tech review completed

## What's changed?

- Alters the `app/jobs/schedule_certificate_completion_job.rb` to take a transition metadata parameter similar to other jobs.
- Implements a rake task to mark teachers as qualified 

- Enter heroku bash console
- Create two CSV files containing just the stem_user_ids for primary and secondary certificates
- Run the rake task on each CSV file with the correct certificate selected.


dcorking (dismissed):
@Caleb-T-Owens I really like this. The narrative nature of the test makes it easy for me to QA without setting up my own local demo of this (so I didn't.). Also great to standardize the interface for passing the meta hash into the jobs.

One suggested patch to the `source` string, that future data archaeologists might benefit from

LGTM 🚂

dcorking:
BTW - don't change this code but worth knowing that a more natural way to process a file with just one value per line is.:
```ruby
File.open("filename") do |file|
  file.each_line do |value|
    puts value
  end
end
```

(`CSV.for_each ... row[0]` is overkill)

dcorking (approved):
🚂 LGTM

PR #1754, opened 2023-08-03
Rescue and report missing course template on cert dashboard

dcorking:
## Status

* Current Status: WIP
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2421

## Review progress:

- [x] Tech review completed

## What's changed?

~Rescue~ Handle and report missing course template on cert dashboard
when recommended course doesn't match a real or stubbed template

This should make dev, review and staging environments more stable, and will also help if there is ever missing data in production Dynamics.

### DIFF
--- a/app/services/achiever/course/template.rb
+++ b/app/services/achiever/course/template.rb
@@ -68,10 +68,27 @@ class Achiever::Course::Template
     templates.map { |course| Achiever::Course::Template.from_resource(course, activities) }
   end
 
+  # @return [Achiever::Course::Template] !!raises ActiveRecord::RecordNotFound!! if none . Use this method if you are using
+  # untrusted data such as a URL path to show a page for a course, otherwise use #maybe_find_by_activity_code.
   def self.find_by_activity_code(activity_code)
     templates = all
     template = templates.find { |val| val.activity_code == activity_code.upcase }
-    raise ActiveRecord::RecordNotFound unless template
+    raise ActiveRecord::RecordNotFound, "Could not find template #{activity_code} in Smart Connector" unless template
+
+    template
+  end
+
+  # @return [Achiever::Course::Template|nil] nil if not found. Silently logs and reports an error if not found, so use this method
+  # if you know the course should exist (because you have a matching course Activity)
+  def self.maybe_find_by_activity_code(activity_code)
+    templates = all
+    template = templates.find { |val| val.activity_code == activity_code.upcase }
+
+    unless template
+      message = "Could not find template #{activity_code} in Smart Connector"
+      Rails.logger.error(message)
+      Sentry.capture_message(message, level: :error)
+    end
 
     template
   end
--- a/app/services/achiever/course/template.rb
+++ b/app/services/achiever/course/template.rb
@@ -1,4 +1,6 @@
 class Achiever::Course::Template
+  include ActionView::Helpers::TextHelper
+
   attr_accessor :activity_code,
                 :age_groups,
                 :booking_url,
@@ -122,5 +124,16 @@ class Achiever::Course::Template
 
   def nearest_occurrence_distance
     occurrences.map(&:distance).compact.min
+  end
+
+  def duration_present?
+    duration_unit.present? && duration_value.present?
+  end
+
+  # For example, "8 hours" or "1 week"
+  def formatted_duration
+    return '' unless duration_present?
+
+    pluralize(duration_value, duration.downcase.chomp('s'))
   end
 end
--- a/app/views/certificates/pathways/_recommended_courses.html.erb
+++ b/app/views/certificates/pathways/_recommended_courses.html.erb
@@ -26,9 +26,9 @@
                 <%= activity_type(pa) %>
               </p>
               <% course = Achiever::Course::Template.maybe_find_by_activity_code(pa.stem_activity_code)%>
-              <% if (duration = course&.duration.presence) && (duration_value = course&.duration_value.presence) %>
+              <% if course&.duration_present? %>
                 <p class="ncce-pathway-courses__location icon-clock govuk-!-margin-top-0">
-                  <%= pluralize(duration_value, duration.downcase.chomp('s')) %>
+                  <%=  course.formatted_duration %>
                 </p>
               <% end %>
               <% if @current_user.achievements.find_by_activity_id(pa.activity_id) %>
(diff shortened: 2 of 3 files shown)


app/views/certificates/pathways/_recommended_courses.html.erb
```diff
@@ -26,8 +26,12 @@
                 <%= activity_type(pa) %>
               </p>
               <p class="ncce-pathway-courses__location icon-clock govuk-!-margin-top-0">
-                <% course = Achiever::Course::Template.find_by_activity_code(pa.stem_activity_code)%>
-                <%= pluralize(course.duration_value, course.duration.downcase.chomp('s')) if course.duration.present? %> 
+                <% begin %>
```
Caleb-T-Owens, line 29: I’m never a huge fan of throwing and catching exceptions as it creates a bit of a nightmare of control flow. In my pathways branch, I’ve added a
```rb
  def self.maybe_find_by_activity_code(activity_code)
    templates = all
    templates.find { |val| val.activity_code == activity_code.upcase }
  end
```
method for potentially finding an activity.
Could you add this instead and handle the nil case?

Unfortunately sending that exception to sentry won’t tell us which CPDs are missing from dynamics because that find_by_activity_code method throws `raise ActiveRecord::RecordNotFound unless template` directly without any message.
What would be quite nice is if we could add a message to that not found, and maybe also add quiet sentry reporting to the maybe_find_by_activity_code method

Caleb-T-Owens (changes requested):

app/views/certificates/pathways/_recommended_courses.html.erb
dcorking: I've done all this, and re-based, except "add a message to that not found" (as I haven't thought of a useful message or a quick way to add it.) Pls re-review.

app/services/achiever/course/template.rb
```diff
@@ -76,6 +77,20 @@ def self.find_by_activity_code(activity_code)
...
+  # @return [Achiever::Course::Template|nil] nil if not found
+  def self.maybe_find_by_activity_code(activity_code)
+    templates = all
+    template = templates.find { |val| val.activity_code == activity_code.upcase }
+
+    return template if template
+
+    message = "Could not find template #{activity_code} in Smart Connector"
+    Rails.logger.error(message)
+    Sentry.capture_message(message, level: :error)
```
Caleb-T-Owens, line 89:
```suggestion
    Sentry.capture_message(message, level: :warn)
```

Given that this should only cause some visual elements to not display, should we put this at warn level rather than error?

app/services/achiever/course/template.rb
```diff
@@ -76,6 +77,20 @@ def self.find_by_activity_code(activity_code)
...
+
+    template
```
Caleb-T-Owens, line 91:
```suggestion
    nil
```

I like the idea of explicitly returning something, but everything in my head is saying it should explicitly be nil as you're returning it early if it has a value.

Or, if you were to restructure the function like:
```rb
# @return [Achiever::Course::Template|nil] nil if not found
  def self.maybe_find_by_activity_code(activity_code)
    templates = all
    template = templates.find { |val| val.activity_code == activity_code.upcase }

    unless template
      message = "Could not find template #{activity_code} in Smart Connector"
      Rails.logger.error(message)
      Sentry.capture_message(message, level: :error)
    end

    template
  end
```

it would be preferable

app/views/certificates/pathways/_recommended_courses.html.erb
```diff
@@ -25,10 +25,12 @@
...
-              <p class="ncce-pathway-courses__location icon-clock govuk-!-margin-top-0">
-                <% course = Achiever::Course::Template.find_by_activity_code(pa.stem_activity_code)%>
-                <%= pluralize(course.duration_value, course.duration.downcase.chomp('s')) if course.duration.present? %> 
-              </p>
+              <% course = Achiever::Course::Template.maybe_find_by_activity_code(pa.stem_activity_code)%>
+              <% if (duration = course&.duration.presence) && (duration_value = course&.duration_value.presence) %>
+                <p class="ncce-pathway-courses__location icon-clock govuk-!-margin-top-0">
+                  <%= pluralize(duration_value, duration.downcase.chomp('s')) %>
+                </p>
+              <% end %>
```
Caleb-T-Owens, line 33: Could we add a method in the template class for `duration_present?` and `formatted_duration` so this code needs less knowledge of a course's internals?

```suggestion
              <% if course.duration_present? %>
                <p class="ncce-pathway-courses__location icon-clock govuk-!-margin-top-0">
                  <%= course.formatted_duration %>
                </p>
              <% end %>
```

app/services/achiever/course/template.rb
```diff
@@ -68,6 +68,7 @@ def self.all
     templates.map { |course| Achiever::Course::Template.from_resource(course, activities) }
   end
 
+  # @return [Achiever::Course::Template] !!raises ActiveRecord::RecordNotFound!! if none
   def self.find_by_activity_code(activity_code)
     templates = all
     template = templates.find { |val| val.activity_code == activity_code.upcase }
```
Caleb-T-Owens, line 74: Can we add a similar message to this raise?

```
raise ActiveRecord::RecordNotFound, "Could not find template #{activity_code} in Smart Connector" unless template
```

Caleb-T-Owens (changes requested):

app/views/certificates/pathways/_recommended_courses.html.erb
Caleb-T-Owens: This is not really in scope so I'm happy for it to be moved to a ticket for a later date

app/services/achiever/course/template.rb
dcorking: I want this to be prominent in Sentry (and Slack us), because the immediate symptom is only visual, but it means a recommended course is actually missing and so the user can't book it.

Caleb-T-Owens: That is reasonable. It might be worth a comment explaining that reasoning or adding it as a note in your commit body

app/views/certificates/pathways/_recommended_courses.html.erb
dcorking: Yes - out-of-scope - (not least because I don't have a fixed IP today to test against Dynamics.)

app/services/achiever/course/template.rb
dcorking: You mean to logs and Sentry? I guess so but we've got on for 4 years without it.

Caleb-T-Owens: I mean to just include it as a message in the error that is already being thrown so we can get some more useful information when we do see the exception. Logging to sentry could also be helpful but generally, that is handled where the exception is caught

Caleb-T-Owens (approved):

PR #1759, opened 2023-08-10
Fix progress bar component

Caleb-T-Owens:
* Before

Before the seeds would not update the programme activity groups in staging or production. There was also an issue with how the last spacing was put in as it made an assumption about how many groupings vs worded groupings there should be. It also is displaying the secondary progress bar when it its not ready yet.
* This commit

I fix the above

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed


dcorking (approved):
LGTM 🚂

PR #1760, opened 2023-08-10
Add + update dynamic links on /i-belong static page

dcorking:
## Status

* Review App link(s): 
* Related to
   -  #1758 
   - https://github.com/NCCE/teachcomputing.org-issues/issues/2425 
   - https://github.com/NCCE/teachcomputing.org-issues/issues/2426

## Review progress:

- [x] Tech review completed

## What's changed?

Add dynamic links to /i-belong static page for logged in and enrolled states

Note that the Champions Info Pack is not ready yet, so I arbitrarily hard-coded a path to the pack for the enrolled state.

Also link to correct dashboard (the certificate itself, not the generic dashboard)

### DIFF
--- a/app/components/non_bordered_cards_component/non_bordered_cards_component.html.erb
+++ b/app/components/non_bordered_cards_component/non_bordered_cards_component.html.erb
@@ -12,7 +12,7 @@
       <% end %>
       <p class="govuk-body">
         <% if card[:link][:link_url].present? %>
-          <%= link_to card[:link][:link_title], card[:link][:link_url], method: (card[:link][:link_method].presence || :get) ,class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
+          <%= link_to card[:link][:link_title], card[:link][:link_url], method: link_method(card) ,class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
             event_action: 'click', event_category: card[:link][:tracking_page], event_label: card[:link][:tracking_label]
           } %>
         <% else %>
--- a/app/components/non_bordered_cards_component/non_bordered_cards_component.html.erb
+++ b/app/components/non_bordered_cards_component/non_bordered_cards_component.html.erb
@@ -12,7 +12,7 @@
       <% end %>
       <p class="govuk-body">
         <% if card[:link][:link_url].present? %>
-          <%= link_to card[:link][:link_title], card[:link][:link_url], class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
+          <%= link_to card[:link][:link_title], card[:link][:link_url], method: (card[:link][:link_method].presence || :get) ,class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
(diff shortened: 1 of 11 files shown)


app/views/pages/i-belong.html.erb
```diff
@@ -36,25 +60,24 @@
...
-                <ol class="govuk-body">
-                  <li>Enrol on the programme</li>
-                  <li>Obtain your free handbook and posters</li>
-                  <li>Complete our ‘Encouraging Girls into GCSE Computer Science’ short course</li>
-                  <li>Participate in a range of recommended activity</li>
-                </ol>
-                <%= link_to 'Enrol on the programme', programme.enrol_path(user_programme_enrolment: { user_id: current_user.id, programme_id: programme.id }), class: 'button govuk-button govuk-!-margin-bottom-0' %>
-              <% end %>
+            <% case state %>
+            <% when :enrolled  %>
```
Caleb-T-Owens, line 64: I'm wondering if in your above when statement, if you could have a variable for `title` `steps` `link_cta` and `link_url` as the template looks identical and then plonk them into just one set of html?

app/views/pages/i-belong.html.erb
```diff
@@ -1,3 +1,27 @@
...
+  <% champion_path = enrol_path %>
+  <% posters_link_title = 'Enrol to request' %>
+  <% posters_path = enrol_path %>
+  <% link_method = :post %>
+<% else %>
+  <% champion_path = login_path %>
+  <% posters_link_title = 'Log in to request' %>
+  <% posters_path = login_path %>
+  <% link_method = :get %>
+<% end %>
```
Caleb-T-Owens, line 23: Firstly, I'm surprised that this doesn't cause an error for being unable to find the `state` method when current_user is nil. I presume the other two variables are not referenced when current_user is nil, but still feels quite fragile.

I'm wondering if for the sake of less mental complexity it would make sense to always define `programme` and the `enrol_path` as the cost of that query is going to be relatively low.


I'm also wondering if we could move these into the controller; I'm not big fan of logic like this sitting in a view, it feels a bit messy.

app/components/non_bordered_cards_component/non_bordered_cards_component.html.erb
```diff
@@ -12,7 +12,7 @@
       <% end %>
       <p class="govuk-body">
         <% if card[:link][:link_url].present? %>
-          <%= link_to card[:link][:link_title], card[:link][:link_url], class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
+          <%= link_to card[:link][:link_title], card[:link][:link_url], method: (card[:link][:link_method].presence || :get) ,class: [{ 'ncce-link': card[:link][:button].blank?, "govuk-button button button--full-width": card[:link][:button].present? }], data: {
```
Caleb-T-Owens, line 15: This looks like what once was a good idea, but now is just too much. Can we define a method in the view component which breaks down these conditionals into more than just one expression?

Caleb-T-Owens (changes requested):
If you've got some time; it would be nice if you could move this page to /pages/enrolment/i-belong-certificate.html.erb and make use of the PagesController#static_programme_page the paradgim of 
```rb
  get '/cs-accelerator', to: 'pages#static_programme_page', as: :cs_accelerator,
                         defaults: { page_slug: 'cs-accelerator' }
```
in the routes file?

Caleb-T-Owens (approved):

PR #1773, opened 2023-08-24
Alter docker configuration to use just one "web" container.

Caleb-T-Owens:
## Status

## Review progress:

- [x] Tech review completed

## What's changed?

Previously, the docker setup used a seperate sidekiq and webpack container. "web" "sidekiq" and "webpack" all use the same dockerfile so it makes no sense to have these running on seperate containers.

This Commit, combines them into just one container called "web". It also adds "teachcomputing.test" into the list of safe redirect URLs to help puma-dev users have a more realistic environment

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -99,4 +99,6 @@ Rails.application.configure do
       exception_object: event.payload[:exception_object] # the exception instance
     }
   end
+
+  config.active_job.queue_adapter = :sidekiq
 end
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,19 +26,6 @@ services:
       - "6379:6379"
     restart: unless-stopped
 
-  sidekiq:
-    build: .
-    command: bundle exec sidekiq
-    volumes:
-      - .:/app
-      - bundle:/usr/local/bundle
-    depends_on:
-      - redis
-      - web
-    env_file:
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -54,14 +41,14 @@ services:
...
     depends_on:
       - db
       - redis
       - docker-host
-      - webpack
     environment:
       - RAILS_DEV_DB_PASS=password
-      - WEBPACKER_DEV_SERVER_HOST=webpack
+      - WEBPACKER_DEV_SERVER_HOST=web
     stdin_open: true
```
dcorking, line 52: I think this might be useful to help your shell do job control
```suggestion
    init: true
    stdin_open: true
```
ref https://docs.docker.com/compose/compose-file/05-services/#init

Caleb-T-Owens: Already there on line 32!

config/environments/development.rb
```diff
@@ -99,4 +99,6 @@
       exception_object: event.payload[:exception_object] # the exception instance
     }
   end
+
+  config.active_job.queue_adapter = :sidekiq
```
dcorking, line 103: 🦅 👁️

dcorking (approved):
LGTM 🚂

PR #1776, opened 2023-08-25
CPD Logic Changes

Caleb-T-Owens:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2438

## Review progress:

- [x] Browser tested
- [x] Tech review completed

## What's changed?

- Adds a new PAG subclass which counts credits as its requirement for completion rather than quantity of achievements
- Refactored PAG superclasss `user_complete?` method to clearer and more concise logic.
- Updated Activity credit counts on activities which are part of primary and secondary

Users are only required to get up to 50 credits (5 hours). This is intentionally softer than the copy which says 6 hours


app/models/programme_activity_grouping.rb
```diff
@@ -15,14 +16,12 @@ def achievements(user)
...
-      completed_activity_count += 1 if completed_activity
-      return true if completed_activity_count >= required_for_completion
-    end
-    nil
+    user_achievement_activity_ids = achievements(user).pluck(:activity_id)
+
+    completed_activity_count = programme_activities
+      .count { _1.activity_id.in? user_achievement_activity_ids }
+
+    completed_activity_count >= required_for_completion
```
dcorking, line 24: Is `#user_complete?` a refactor with no functional changes? If not what changed here?

dcorking (approved):
Nicely done LGTM. One question (below).
🚂

app/models/programme_activity_grouping.rb
Caleb-T-Owens: It's a refactor with next to no functional changes. It returns false instead of nil when It's "not true". This doesn't affect any of the locations in which it is used though

PR #1784, opened 2023-09-05
modifies cta for ibelong handbook

nyashamaphosa:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2459
* Review app https://teachcomputing-pr-1784.herokuapp.com/i-belong

## What's changed?

Handbook link and phrase changes according to user session state


Caleb-T-Owens (changes requested):
From Slack:
Really good idea moving the logic into a helper, I like it :+1:.

For this one page, we do however have a switch statement which handles not logged in, unenrolled, and enrolled in the pages_controller. Could you instead have a “handbook_download_url” and “handbook_download_title” variables assigned in there following the existing pattern?

nyashamaphosa:
> From Slack: Really good idea moving the logic into a helper, I like it 👍.
> 
> For this one page, we do however have a switch statement which handles not logged in, unenrolled, and enrolled in the pages_controller. Could you instead have a “handbook_download_url” and “handbook_download_title” variables assigned in there following the existing pattern?

resolved thank you

app/controllers/pages_controller.rb
```diff
@@ -36,18 +36,24 @@ def i_belong
       posters_link_method = :get
       cta_link_path = i_belong_path
       cta_link_method = :get
+      handbook_download_url = 'https://static.teachcomputing.org/I+Belong+Handbook.pdf'
```
Caleb-T-Owens, line 39: Can you use the helper `helpers.i_belong_handbook_url` (defined in S3LinkHelper) rather than specifying the URL again here.

Caleb-T-Owens (changes requested):
Great stuff using the same when statement, I forgot about this helper method when doing the original review, could you quickly make the change over

app/controllers/pages_controller.rb
nyashamaphosa: resolved thank you

Caleb-T-Owens (approved):

PR #1801, opened 2023-09-18
seeds courses

nyashamaphosa:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2476

## What's changed?

Seeds courses


Caleb-T-Owens (changes requested):
Could you add in the quantity of credits for these activities

nyashamaphosa:
> Could you add in the quantity of credits for these activities

resolved thank you

Caleb-T-Owens:
@nyashamaphosa I've had a discussion with Dapo about these courses and we're not sure that they're available just yet, so we're going to want to just hold off until we've got a better idea of how these multi-part CPDs should behave from Dynamics

Caleb-T-Owens:
See ticket for closure reason

PR #1802, opened 2023-09-18
Removing old programme activities from primary.

Caleb-T-Owens:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2475

## Review progress:

- [x] Browser tested
- [x] Tech review completed

## What's changed?

- Removes old programme activities. This will mean users won't be able to see them and completed achievements for those activities won't count towards certificates.
- Added a flag to show a message to users indicating that their pathway has been changed

- Run `rake migrate_old_primary_pathways`
- Remove legacy primary pathways
- Remove the rake task


dcorking:
- Can we make a note to delete `migrate_old_primary_pathways.rb` after this is live, please?
- Shall we commit the updated `erd.pdf`? (if there is a conflict then just run db:migrate on main after merging this, and push that commit straight to main.)

Caleb-T-Owens:
@dcorking Note made, I'm going to locally resolve the conflicts

lib/tasks/migrate_old_primary_pathways.rake
```diff
@@ -0,0 +1,15 @@
+desc 'Takes users in the old primary pathways and moves them to new ones'
+task migrate_old_primary_pathways: :environment do
+  old_developing = Pathway.find_by(slug: 'developing-in-the-classroom')
+  new_developing = Pathway.find_by(slug: 'developing-in-the-classroom-2')
+  old_specialising = Pathway.find_by(slug: 'specialising-or-leading')
+  new_specialising = Pathway.find_by(slug: 'specialising-or-leading-2')
+
+  UserProgrammeEnrolment
+    .where(pathway: old_developing)
+    .update(pathway_id: new_developing.id, message_flags: { primary_pathway_migrated: true })
```
dcorking, line 10: Did you mean?
```suggestion
    .update_all(pathway_id: new_developing.id, message_flags: { primary_pathway_migrated: true })
```
`#update` is an undocumented public method. I may be wrong here.

lib/tasks/migrate_old_primary_pathways.rake
```diff
@@ -0,0 +1,15 @@
...
+
+  UserProgrammeEnrolment
+    .where(pathway: old_specialising)
+    .update(pathway_id: new_specialising.id, message_flags: { primary_pathway_migrated: true })
```
dcorking, line 14: ditto
```suggestion
  UserProgrammeEnrolment
    .where(pathway: old_specialising)
    .update_all(pathway_id: new_specialising.id, message_flags: { primary_pathway_migrated: true })
```

dcorking (commented):
Clean, thorough and well tested. I'd like your thoughts on the following before you ship this, please.

As this is 'modal' it feels quite intrusive. Did you and Dapo discuss a non-modal popup? So that users can do something else before they read and acknowledge the popup. I am a little concerned that they might dismiss it in frustration before they've read it.

I noticed a small nit while testing:
```
 message_flags: {"primary_pathway_migrated"=>nil}>
```
between deploying the code and running the rake task, my user was in the above state: that meant they got to see and acknowledge the modal, but they hadn't yet been migrated. They saw the modal again.
YMMV as to whether it is worth fixing this, or cleaner just to put the app in maintenance mode for the minute or two that it might take for you to fire up a `heroku run` dyno and complete the rake task, or even doing nothing because a tiny number of people will open their dashboard during this window.

db/seeds/programme_activity_groupings_helper.rb
```diff
@@ -4,3 +4,9 @@ def maybe_attach_activity_to_grouping(grouping, activity_slug, order, legacy: fa
     programme_activity.update(programme_activity_grouping_id: grouping.id, order:, legacy:)
   end
 end
+
+def maybe_detach_activity_from_grouping(grouping, activity_slug)
+  if activity = Activity.find_by(slug: activity_slug)
+    grouping.programme.programme_activities.find_by(activity_id: activity.id)&.destroy
+  end
+end
```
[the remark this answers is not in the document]
Caleb-T-Owens: This method is detaching an activity from a grouping. With how they're attached to the PAG, it doesn't matter if we search through all the programme's programme activities, but it wouldn't hurt if I added in a little bit of extra constraint.

lib/tasks/migrate_old_primary_pathways.rake
Caleb-T-Owens: update is the same as update on a single record. The potential issue with .update() is that it runs validations and hooks whereas update_all doesn't. update_all() doesn't support getters and setters, IE .update_all(pathway: foo) would fail as "pathway" isn't a column.

I originally wrote it with .update and changed over to passing pathway_id and manually passing the hash so that I could use update_all and it seems like I just forgot to change the method I was calling!

Caleb-T-Owens:
I thought about the fact that dismissing the popup would put ` message_flags: {"primary_pathway_migrated"=>nil}` in the database, but I stuck with `enrolment.update(message_flags_primary_pathway_migrated: nil)` for two reasons:
1. Its a bit tidier than doing something like:
```
enrolment.message_flags.delete! :primary_pathway_migrated
unless enrolment.save
```
2. It could be worthwhile information being able to query how many users have closed the modal
3. If we call `enrolment.message_flags_primary_pathway_migrated` it will return nil either way.

dcorking (approved):
LGTM. I suggest you merge as soon as Dapo is happy 🚂

PR #1812, opened 2023-10-02
Made achievements global

Caleb-T-Owens:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

This PR makes achievements global; completing an activity that is in multiple programmes will complete it everywhere


app/controllers/achievements_controller.rb
```diff
@@ -14,12 +14,7 @@ def create
 
       @achievement.transition_to(:complete, metadata)
 
-      if @achievement.programme
-        case @achievement.programme.slug
-        when 'i-belong', 'primary-certificate', 'secondary-certificate'
-          CertificatePendingTransitionJob.perform_now(@achievement.programme, current_user.id, source: 'AchievementsController.create')
-        end
-      end
+      CertificatePendingTransitionJob.perform_now(current_user, { source: 'AchievementsController.create' })
```
markjs, line 17: It's common practice to omit braces when passing arguments like this. The code being replaced here omits them so my assumption would be it's standard within this project to omit them too.

markjs: `CertificatePendingTransitionJob.perform_now(current_user, source: 'AchievementsController.create')`

app/models/achievement.rb
```diff
@@ -52,21 +44,8 @@ class Achievement < ApplicationRecord
...
-  end
-
-  def issue_badge
-    return unless programme&.badgeable? && programme&.user_enrolled?(user)
-    return unless programme.badges.active.first
-    return if user_has_badge?
-
-    Credly::IssueBadgeJob.perform_later(user.id, programme.id) if first_stem_achievement?
-  end
+  scope :belongs_to_programme, ->(programme) { joins(activity: { programme_activities: :programme }).where(activities: { programme_activities: { programme: }}) }
```
markjs, line 47: I think this name would read better as `belonging_to_programme` since it's acting on a plural object. Same for the one below.

app/controllers/achievements_controller.rb
Caleb-T-Owens: I'm specifically passing in the hash as the second positional argument. The def perform(...) has kwargs meaning that if I passed just `source: '...'` it would throw an error due to a non-existent kwarg

app/models/achievement.rb
Caleb-T-Owens: Yep - I was reading it as "where belongs_to_programme" in my head, but belonging_to_programme works far better stand-alone

app/controllers/api/achievements_controller.rb
```diff
@@ -6,12 +6,9 @@ def create
...
       if achievement.save
-        case achievement.programme&.slug
-        when 'cs-accelerator'
-          AssessmentEligibilityJob.perform_now(user.id, nil)
-        when 'primary-certificate'
-          CertificatePendingTransitionJob.perform_now(achievement.programme, user.id, nil)
-        end
+        AssessmentEligibilityJob.perform_now(user.id)
+        CertificatePendingTransitionJob.perform_now(user, { source: 'AdminAchievementsController.create' })
+
```
Caleb-T-Owens, line 11: Note: As achievements are now global, they may cause any of the certificates to be completed so we need to call both the AssessmentEligibilityJob for CSA, and the modified CertificatePendingTransitionJob which now has an internal loop which runs on Primary, Secondary, and I Belong

app/controllers/achievements_controller.rb
markjs: Well I stand corrected

app/controllers/api/achievements_controller.rb
markjs: How do you feel about moving the calling of these two jobs into a method on the user? Might be nice for this line to be `user.transition_certificates_if_eligible` or similar rather than needing to call these two jobs in a row without an obvious reason in code for why. I suppose the source still needs passing through perhaps? Feels like this is a part of a refactor when we could go a lot further to add a bit more meaning to what's going on where.

Caleb-T-Owens: The `AssessmentEligibilityJob` is specifically for CSA and its job is somewhat distinct from the `CertificatePendingTransitionJob`. The `AssessmentEligibilityJob` checks to see if enough activities have been completed for the user to take the CSA programmes test and if they have, it will send them an email letting them know they're eligible. The `CertificatePendingTransitionJob` will transition a programme to `:pending` if the user has completed adequate activities.

There are also examples where we actually want to call the two separately, [namely after the CSA test has been completed](https://github.com/NCCE/teachcomputing.org/blob/3d9237cda701c50fed3eca36cd2170f02d183111/app/jobs/cs_accelerator_enrolment_transition_job.rb). This is because we need to test and see whether the Secondary programme has been completed as completing CSA is a requirement for completing Secondary.

I think, even if we did always want them to be called together, I'm not convinced that putting them in a method together in the user model is a helpful abstraction. In terms of reading the code, I can quite clearly see that two jobs have been started here. If they were aliased via a method, I'd need to maintain that mental overhead. I'm not entirely convinced that it is the user's responsibility either

If we still needed that large case statement as a common pattern in multiple places, I would agree that it would be a good idea to try and concentrate that in a common place, but as it currently is, we're going to just end up aliasing two direct calls.

~~

The time when we need to make these two calls is when an achievement becomes complete. This is because our completion requirements are based on finishing activities, which are either the community or CPD/PDL ones.

What could be a useful refactor is moving the two calls from each instance in the achievement_controllers and putting them into the hook in the state machine which is fired when an activity transitions to complete. How does that sound to you?

markjs: If the state machine already exists and tying the state directly to that is definitely correct in all cases then that seems like a good place to put it.

markjs (approved):

app/controllers/api/achievements_controller.rb
Caleb-T-Owens: We ended up not moving it into the state machine as there was one point in the app where it would have caused trouble as achievements were marked complete in a hot loop which would trigger the job many times.

PR #1815, opened 2023-10-05
updates primary and secondary pathways

nyashamaphosa:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2486

## What's changed?

*Updates to primary and secondary pathways*


Caleb-T-Owens (changes requested):
Unfortunately this change is not sound.

You've created a second activity with a slug which is the same as another one. While there is no database constraint for their uniqueness there are multiple points in the code which find_by(slug:) and as such would fail to find the secondary activity.

After you change the activity slug, you'll also need to update the secondary pathways and programme activity groupings to accommodate

Caleb-T-Owens:
These changes have been superseded in another PR

PR #1823, opened 2023-10-17
Now pulling school name from oauth

Caleb-T-Owens:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2457

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

We pull their name from their login details.

We ended up not pulling names from the identity server at certificate transition as it was too complex. Users need to log in when their session is invalidated which is generally each time they close their browser (if not more often then that). Because of this, only a very small minority of people will ever see the modal to put in their school name.


app/controllers/user_controller.rb
```diff
@@ -29,8 +29,6 @@ def trn_redirect_path
     end
 
     def user_params
-      params.require(:user).permit(
-        :school_name
-      )
+      params.require(:user).permit()
```
markjs, line 32: What does `permit` without any arguments achieve here? My understanding was `permit!` would permit anything but this permits nothing? I may be wrong.

app/controllers/user_programme_enrolments_controller.rb
```diff
@@ -14,6 +14,18 @@ def create
     end
   end
 
+  def update
+    user_programme_enrolment = current_user.user_programme_enrolments.find(params[:id])
+
+    if user_programme_enrolment.update(user_prgoramme_enrolment_update_params)
```
markjs, line 20: It's a good job this method name was typo'd in both places so it still worked. Let's maybe spell programme (not prgoramme) correctly though 😂

markjs (approved):
Looks much better. Nice one 👍 

(One question and one typo but happy to approve and you can deal with answering/resolving those)

markjs (approved):

PR #1826, opened 2023-10-18
Added CSA to secondary computing dashboard.

Caleb-T-Owens:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2496 https://github.com/NCCE/teachcomputing.org-issues/issues/2492

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

This PR adds the CSA as an item in the progress bar at the top of the secondary programme dashboard.

We've added a handful of new programmes to T/C now and I've held off "drying" up the code as each of the programmes has had a quite distinct set of requirements for completion. In this PR, I've started an abstraction which should let us get rid of having multiple programme controllers, and move to having just one generic controller in the future.

It's clear that we have many different completion requirements for each of our programmes, some being the number of activities needing to be completed, others requiring other programmes to be completed first, and others requiring quizzes to be completed.

I've started on a new abstraction for "ProgrammeObjective"s. The ProgrammeObjective role has been designed to conform to part of the existing interface of the ProgrammeActivityGroupings. Primarily the "user_complete?(user)" method; thus with a little tweaking, making ProgrammeActivityGroupings a programme objective.

I've changed the old logic in the base Programme class which iterated over all the ProgrammeActivityGroupings, and now ask it to iterate over all the ProgrammeObjectives. For most Programmes the list of ProgrammeObjectives will the the same as the ProgrammeActivityGroupings, but it allows us to create different objectives that can test different actions the user has made.

ProgrammeObjectives also have a "objective_displayed_in?(location)" method. Currently, we have two places where we want to display the progress to users. The top "progress bar" and in the main "body" of the programme dashboards. This PR has modified the ProgressBarComponent to use the list of ProgrammeObjectives which are displayed in the :progress_bar, allowing in this case for the CSA programme completion to be one of the steps that needs completing.

In the future, we will be able to create a new generic programme controller which similarly iterates all the ProgrammeObjectives that belong to the body, rendering a ViewComponent which is provided via a method on the ProgrammeObjectives playing the body role, and dynamically render the relevant UI elements.

### DIFF
--- a/app/models/programme_activity_grouping.rb
+++ b/app/models/programme_activity_grouping.rb
@@ -71,8 +71,12 @@ class ProgrammeActivityGrouping < ApplicationRecord
     completed_legacy_activities + completed_non_legacy_activities + non_completed_non_legacy_activities
   end
 
-  def objective_displayed_in?(role)
-    role.in? %i[progress_bar body]
+  def objective_displayed_in_progress_bar?
+    true
+  end
+
+  def objective_displayed_in_body?
+    true
   end
 
   def progress_bar_path
--- a/app/models/programme_activity_grouping.rb
+++ b/app/models/programme_activity_grouping.rb
@@ -1,3 +1,4 @@
+# Plays "ProgrammeObjective" role app/lib/programme_objectives/readme.md
 class ProgrammeActivityGrouping < ApplicationRecord
   include ActionView::Helpers::TagHelper
   include StiPreload
@@ -68,5 +69,13 @@ class ProgrammeActivityGrouping < ApplicationRecord
       .select { completed_activity_ids.include?(_1.activity_id) }
 
     completed_legacy_activities + completed_non_legacy_activities + non_completed_non_legacy_activities
+  end
+
+  def objective_displayed_in?(role)
+    role.in? %i[progress_bar body]
+  end
+
+  def progress_bar_path
+    "##{id}"
   end
 end
--- a/app/models/programmes/secondary_certificate.rb
+++ b/app/models/programmes/secondary_certificate.rb
@@ -14,12 +14,6 @@ module Programmes
       SecondaryMailer
     end
 
-    def user_meets_completion_requirement?(user)
-      return false unless Programme.cs_accelerator.user_completed?(user)
-
-      super(user)
-    end
-
     def path
       secondary_certificate_path
     end
@@ -46,6 +40,17 @@ module Programmes
 
     def enrichment_enabled?
       true
+    end
+
+    def programme_objectives
+      [
+        ProgrammeObjectives::ProgrammeCompletionRequired.new(
+          required_programme: Programme.cs_accelerator,
+          progress_bar_title: 'Complete the Secondary subject knowledge',
+          progress_bar_path: cs_accelerator_path
+        ),
+        *programme_activity_groupings.includes(:programme_activities).order(:sort_key)
+      ]
     end
   end
 end
--- a/app/views/certificates/i_belong/_achievement_list.html.erb
+++ b/app/views/certificates/i_belong/_achievement_list.html.erb
@@ -1,4 +1,4 @@
-<p class="govuk-body ncce-activity-list__title govuk-!-margin-top-8 <%= add_groups_complete_icon_class(current_user, @professional_development_groups) %>">
+<p id="<%= @professional_development_groups.first&.id %>" class="govuk-body ncce-activity-list__title govuk-!-margin-top-8 <%= add_groups_complete_icon_class(current_user, @professional_development_groups) %>">
   <%= t('.development.title.html') %>
 </p>
 <ul class="govuk-list ncce-activity-list ncce-activity-list--programme" role="list">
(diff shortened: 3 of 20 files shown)


spec/requests/certificates/secondary_certificate/show_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe Certificates::SecondaryCertificateController do
   let(:user) { create(:user) }
+  let!(:cs_accelerator) { create(:cs_accelerator) }
```
Caleb-T-Owens, line 5: Secondary has a programme_objective which requires CSAccelerator programme to be present. Ideally we should have a better way of seeding the programmes like they are in production in our tests. Without putting together that system, I think its adequate to just create cs_accelerator manually a head of time.

spec/models/programme_spec.rb
```diff
@@ -318,4 +318,16 @@
       expect { generic_programme.short_name }.to raise_error(NotImplementedError)
     end
   end
+
+  describe '#mailer' do
```
Caleb-T-Owens, line 322: These were just some missing tests in some of the files that I altered so I've added them in here to take it up to 100% coverage

spec/components/progress_bar_component_spec.rb
```diff
@@ -4,11 +4,10 @@
     let(:user) { create(:user) }
     let(:primary_certificate) { create(:primary_certificate) }
     let(:primary_programme_activity_groupings) do
-      create_list(:programme_activity_grouping, 4, programme: primary_certificate).tap do |groupings|
-        groupings[0].update(progress_bar_title: nil)
-        groupings[1].update(progress_bar_title: 'Complete professional development')
-        groupings[2].update(progress_bar_title: 'Develop your teaching practice')
-        groupings[3].update(progress_bar_title: 'Develop computing in your community')
+      create_list(:programme_activity_grouping, 3, programme: primary_certificate).tap do |groupings|
+        groupings[0].update(progress_bar_title: 'Complete professional development')
```
Caleb-T-Owens, line 8: Having programme_activity_groupings without titles was a legacy idea and no longer exists in the application so its safe to remove that part of the test

app/views/certificates/i_belong/_achievement_list.html.erb
```diff
@@ -1,4 +1,4 @@
-<p class="govuk-body ncce-activity-list__title govuk-!-margin-top-8 <%= add_groups_complete_icon_class(current_user, @professional_development_groups) %>">
+<p id="<%= @professional_development_groups.first&.id %>" class="govuk-body ncce-activity-list__title govuk-!-margin-top-8 <%= add_groups_complete_icon_class(current_user, @professional_development_groups) %>">
```
Caleb-T-Owens, line 1: In production and dev-seeded environments, IB, Primary, and Secondary all have just one professional development group so its safe to take just the first one's ID.

We are adding these IDs in so we can hyperlink to different sections of the page in the Progress Bar

The reason I'm using `&.` is that some of our tests don't fully construct all of the programme activity groupings. This is generally just a greater issue with how the supporting data in our tests is put together and not something that I think is justified in being solved here

app/models/programmes/secondary_certificate.rb
```diff
@@ -14,12 +14,6 @@ def mailer
       SecondaryMailer
     end
 
-    def user_meets_completion_requirement?(user)
```
Caleb-T-Owens, line 17: We can safely remove this as the responsibility of checking that the user has completed CS accelerator has been moved into a ProgrammeObjective

app/models/programme_activity_grouping.rb
```diff
@@ -69,4 +70,12 @@ def order_programme_activities_for_user(user, pathway)
 
     completed_legacy_activities + completed_non_legacy_activities + non_completed_non_legacy_activities
   end
+
+  def objective_displayed_in?(role)
+    role.in? %i[progress_bar body]
+  end
+
+  def progress_bar_path
```
Caleb-T-Owens, line 78: The progress_bar_path and its corresponding target in the DOM is a private contract of this object.

If it's rendered in the progress_bar, it should also be rendered in the body which will have the matching ID on an element

app/views/certificates/i_belong/_achievement_list.html.erb
markjs: If there's only one, could the controller just set `@professional_development_group` and be responsible for the `first` logic instead of it being a bit tucked away in this view?

markjs: Or could the association even be updated so there's only ever one? Little bit smelly to have a data model that allows multiple of something where only one is relevant and used by the code.

markjs (changes requested):
Have added a few small comments.

I'm also struggling to wrap my head around the various `displayed_in?` methods which can take `:progress_bar` or `:body` as arguments. That feels a little bit magic symbol to me. Are `progress_bar` and `body` the only valid options at this stage? Is there a time where this number of options is increasing significantly? I'm wondering if this could be simplified a bit and we just have methods which explicitly reference `body` or `progress_bar` in their names and we don't try and be too fancy with it.

PR #1873, opened 2023-12-19
Digital Summative Assessment Button

msquance-stem:
## Status

* Closes: https://github.com/NCCE/teachcomputing.org-issues/issues/2530
* 
## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Added the new card for external links to the unit page, this has been done as a mixed component that supports files and external links so ordering can be changed easily. Created a component to support the external link in the style of a file card so they look similar. Also made the changes to the GRAPHQL query to pull the new digital summative assessment from curriculum.

### DIFF
--- a/app/components/mixed_cards_component.rb
+++ b/app/components/mixed_cards_component.rb
@@ -1,12 +1,12 @@
 # frozen_string_literal: true
 
 class MixedCardsComponent < ViewComponent::Base
-  def initialize(card_groups:)
-    @card_groups = card_groups
+  def initialize(cards:)
+    @cards = cards
   end
 
   def render?
-    @card_groups.present?
+    @cards.present?
   end
 end
 
--- a/app/components/mixed_cards_component.rb
+++ b/app/components/mixed_cards_component.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class MixedCardsComponent < ViewComponent::Base
+  def initialize(card_groups:)
+    @card_groups = card_groups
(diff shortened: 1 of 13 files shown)


app/components/mixed_cards_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class MixedCardsComponent < ViewComponent::Base
```
Caleb-T-Owens, line 3: Would it be better to just have one array which is `cards` which contains objects like:
```rb
{
  type: :file,
  details: { ... }
}
```

So you can then iterate over them all like:
```
<% @cards.each do |card| %>
  <% case card_group[:type] %>
    <% when :file %>
      <%= render FileCardComponent.new(card[:details], title: @title, tracking: @tracking) %>
    <% when :external_link %>
      <%= render ExternalLinkCardComponent.new(card[:details]) %>
    <% end %>
  <% end %>
<% end %>
```

Meaning that we're not restricted by any ordering so if we so desired the external_link could be in the middle of the files

Caleb-T-Owens (commented):
Other than a minor comment, looking pretty good

app/components/mixed_cards_component.rb
msquance-stem: Yeah its probably clearer that way. Will make the change

Caleb-T-Owens (approved):
LGTM 🚀

PR #1877, opened 2023-12-21
Requested text changes to certificates

msquance-stem:
## Status

* Resolves https://github.com/NCCE/teachcomputing.org-issues/issues/2545


msquance-stem (commented):
Changes double checked

PR #1882, opened 2024-01-04
Creating interfaces for I-Belong Rollbacks through administrate

msquance-stem:
Updated the administrate interfaces to allow the couple of tasks required for I-Belong rollbacks to be done via administrate, rather than having to use the console. Also added testing to cover administrate.

## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2551

## Review progress:

- [ ] Tech review completed

## What's changed?

- Destroy enabled on achievements in administrate
- Added new field to administrate that works on StateMachines, to allow for status changes
- Removed the filter in coverage to exclude administrate sections
- Added new category for components in coverage
- Created tests to cover all the administrate controllers


Caleb-T-Owens:
@msquance-stem Apologies, it seems like there is now a conflict due to the addition of standardrb

app/dashboards/achievement_dashboard.rb
```diff
@@ -12,9 +12,10 @@ class AchievementDashboard < Administrate::BaseDashboard
     user: Field::BelongsTo,
     supporting_evidence_attachment: Field::HasOne,
     supporting_evidence_blob: Field::HasOne,
+    self_verification_info: Field::String,
     current_state: Field::String.with_options(searchable: false),
     id: Field::String,
-    created_at: Field::DateTime,
+    created_at: Field::DateTime.with_options(format: "%d/%m/%Y %H:%M"),
```
Caleb-T-Owens, line 18: Why do we specify a format only here?

If we're going to use a particular format, could we use it consistently throughout the admin panel.

Presumably you could define a constant `FORMATTED_DATE_TIME = Field::DateTime.with_options(format: "%d/%m/%Y %H:%M")` in the base dashboard

app/dashboards/user_programme_enrolment_dashboard.rb
```diff
@@ -28,7 +28,7 @@ class UserProgrammeEnrolmentDashboard < Administrate::BaseDashboard
     user
     programme
     pathway
-    current_state
+    state_machine
```
Caleb-T-Owens, line 31: I'm not sure how I feel about calling it state_machine. The idea that its a state machine is an implementation detail. The admin user should be looking at the "current_state" which may or may not be powered by a state machine.

app/controllers/admin/user_programme_enrolments_controller.rb
```diff
@@ -42,5 +42,17 @@ class UserProgrammeEnrolmentsController < Admin::ApplicationController
...
+      p = params.require(:user_programme_enrolment).permit(:state_machine)
+      user_programme_enrolment = UserProgrammeEnrolment.find(params[:id])
+      result = user_programme_enrolment.transition_to(p[:state_machine].to_sym)
+      if result
+        flash[:notice] = "State changed to #{user_programme_enrolment.current_state} for #{user_programme_enrolment.programme.title}"
+      else
+        flash[:alert] = "Unable to change state"
+      end
+      redirect_to controller: :users, action: :show, id: user_programme_enrolment.user.id
+    end
```
Caleb-T-Owens, line 56:
```suggestion
    def update
      user_programme_enrolment = UserProgrammeEnrolment.find(params[:id])

      if user_programme_enrolment.transition_to(user_programme_enrolment_params[:current_state].to_sym)
        flash[:notice] = "State changed to #{user_programme_enrolment.current_state} for #{user_programme_enrolment.programme.title}"

        redirect_to controller: :users, action: :show, id: user_programme_enrolment.user.id
      else
        flash[:alert] = "Unable to change state"

        redirect_to controller: :users, action: :edit, id: user_programme_enrolment.user.id
      end
    end

    private

    def user_programme_enrolment_params
      params.require(:user_programme_enrolment)
        .permit(:current_state)
    end
```

Caleb-T-Owens (changes requested):

app/dashboards/achievement_dashboard.rb
msquance-stem: Good point, I did it on this one to make the table more readable, but we should be consistent, I have created a base dashboard and put in the suggested constant and updated the rest of the dashboards

app/models/state_machines/assessment_attempt_state_machine.rb
```diff
@@ -12,7 +12,7 @@ class StateMachines::AssessmentAttemptStateMachine
   transition from: STATE_COMMENCED, to: STATE_PASSED
   transition from: STATE_COMMENCED, to: STATE_FAILED
 
-  after_transition(to: :failed) do |assessment_attempt, _transition|
+  after_transition(to: :passed) do |assessment_attempt, _transition|
```
Caleb-T-Owens, line 15: Just highlighting this for future reference; This might have been causing some of the confusion around the credly badges.

The code was firing when a level and CSA were failed rather than passed.

Caleb-T-Owens (approved):
LGTM 🚀 Thanks for those changes 👍

PR #1901, opened 2024-01-16
Updating programme mapping to allow for new tag names

msquance-stem:
## Status

* Current Status: Ready for Tech/Front end review
* Review App link(s): https://teachcomputing-pr-1901.herokuapp.com/
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2569

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Created new mapping of dynamic programme names to programme slugs
Updated by certificate filtering
Adding certificate names to programmes
Unifying the tag code into one helper method
Updates of scss class names

### DIFF
--- a/app/components/course_component.rb
+++ b/app/components/course_component.rb
@@ -2,21 +2,17 @@
 
 class CourseComponent < ViewComponent::Base
   include ViewComponent::Translatable
-  include ProgrammesHelper
 
   delegate :stripped_summary,
     :course_meta_icon_class,
     :course_type,
     :view_course_phrase,
     :tracking_data,
+    :display_programme_tag,
     to: :helpers
 
   def initialize(course:, filter:)
     @course = course
     @course_filter = filter
-  end
-
-  def get_course_tag(programme)
-    display_programme_tag programme
   end
 end
--- a/app/components/course_component.rb
+++ b/app/components/course_component.rb
@@ -2,6 +2,7 @@
 
 class CourseComponent < ViewComponent::Base
   include ViewComponent::Translatable
+  include ProgrammesHelper
 
   delegate :stripped_summary,
     :course_meta_icon_class,
@@ -16,8 +17,6 @@ class CourseComponent < ViewComponent::Base
   end
 
   def get_course_tag(programme)
-    return "#{programme} certificate" if %w[Primary Secondary].include?(programme)
-
-    programme
+    display_programme_tag programme
   end
 end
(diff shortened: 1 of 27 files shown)


app/models/programme.rb
```diff
@@ -45,6 +45,9 @@ def short_name
     raise NotImplementedError
   end
 
+  def certificate_name
+  end
```
Caleb-T-Owens, line 49: This ought to raise NotImplementedError, as otherwise we could end up with a slug being displayed as ""

app/services/achiever/course/template.rb
```diff
@@ -156,18 +160,7 @@ def self.find_many_by_stem_course_template_nos(stem_course_template_nos)
...
-    when "secondary-certificate"
-      @programmes.include?("Secondary")
-    when "primary-certificate"
-      @programmes.include?("Primary")
-    when "i-belong"
-      @programmes.include?("I Belong")
-    when "a-level-certificate"
-      @programmes.include?("A Level")
-    end
+    @programmes.include?(certificate)
```
Caleb-T-Owens, line 163: Does this need to go through the mapping?

app/components/course_component.rb
```diff
@@ -16,8 +17,6 @@ def initialize(course:, filter:)
   end
 
   def get_course_tag(programme)
-    return "#{programme} certificate" if %w[Primary Secondary].include?(programme)
-
-    programme
+    display_programme_tag programme
```
Caleb-T-Owens, line 20: Could you add :display_programme_tag to the "delegate" call to helpers, rather than including all of the ProgrammesHelper helpers

app/services/achiever/course/template.rb
Caleb-T-Owens: Ahhh! no, I see what is going on. You've mapped from the Title Case to slugs right at the very start. Nice

msquance-stem: No, as we are remapping the programme code from dynamic to the slug, and the filter on the course page passes the slug as its value, so there is no need to map backwards

Caleb-T-Owens (approved):
Looks like a good improvement 👍. I'm happy for it to go live, but there are two minor comments that would be nice if you have time

Taking a look, we will only have one other instance of the short_name method in searchable_site_pages.rb:60, could we change that over to using the new certificate_name method there, and then take out all of the short_name.

Could you also add in some code tests for the new certificate_name method. I'd like to keep coverage as near to 100% as possible

Caleb-T-Owens (approved):
Looks good to me! A pass from Hannah and we're off to the races

hannah-storm (approved):
All looks good to me!

PR #1902, opened 2024-01-17
Update seeds to have correct counts of credits

Caleb-T-Owens:
Also changed to be more noisy when a save fails

## Status

* Review App link(s): 

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Updates all Primary/Secondary/IBelong/ALevel CPDs to have the correct number of credits.

Also calls .save! rather than .save so we get some feedback if a record fails to update


msquance-stem (approved):
LGTM

PR #1905, opened 2024-01-18
Improve error handling for curriculum

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2567

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Added rescue for FaradayServerErrors
Made both error types report to sentry

## Reason for change

There was an attempt to inject code into the URL of the curriculum pages, these injections caused the Curriculum GraphQL server to respond with Graphlient::Errors::FaradayServerError. We have decided to have those errors instead return a 404 error rather than passing down the 500 to the end user. These errors will be logged to Sentry so we can continue to monitor for any future issues.

### DIFF
--- a/app/services/curriculum_client/request.rb
+++ b/app/services/curriculum_client/request.rb
@@ -20,6 +20,10 @@ module CurriculumClient
       raise ActionController::RoutingError, e.message if e.message.include?("not found")
 
       raise
+    rescue Graphlient::Errors::FaradayServerError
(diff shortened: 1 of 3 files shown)


Caleb-T-Owens (approved):
Hi, this code looks good; could you just write in your commit message or PR body an example of the effect so future devs can have a better understanding as to why we're repackaging these erorrs

PR #1908, opened 2024-01-19
2565/report generation

Caleb-T-Owens:
## Status

* Review App link(s):
* Related to https://github.com/NCCE/teachcomputing.org-issues/issues/2565

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Added code for report generation
- This is ONLY the report generation for the users; the achievement data can be directly pulled out of the database with relative ease in Looker Studio.
- I've added two methods in `ProgrammeActivityGroupingCompletion` with the purpose behind them having a way of checking whether those particular programme objectives have been completed in bulk, which enables us to use the same logic between report generation and regular account checking
- I have tested on a local copy of the production database and it runs in a little over a minute and creates ~39k records


msquance-stem (approved):
LGTM

PR #1914, opened 2024-01-29
Fixing issues in badge assignment

msquance-stem:
## Status

* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2583

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Changed the API call to credly, the email version seems to give inconsistent results, current working theory is it takes time for that search to become active as using email appears to initially fail, but then later returns result. Using the issuer_earner_id seems to not have the same issue. I believe this should help resolve the 422 errors we get when issuing badges, which might also account for the small number of missing badges.

I swapped the ordering of checks in the IssueBadgeJob, so that it checks locally if the user meets the badge requirements before it checks credly, seems more logical to check DB before making API call.

Found a bug in the cs_accelerator programme where due to it not having programme activity groups it was returning true for user_meets_completion_requirement meaning that you could get the badge for just doing any activity (e.g online only). I have added programme_objectives to the programme that check if the user has completed the assessment to stop it resolving only to true. I have added extra testing to ensure that the credly badge checker is correct for different styles of programmes.

Created a job that returns an array of any missing badges for a given number of days, to allow us to check for issues.

### DIFF
--- a/app/jobs/badge_assignment_check_job.rb
+++ b/app/jobs/badge_assignment_check_job.rb
@@ -2,26 +2,6 @@ class BadgeAssignmentCheck < ApplicationJob
   queue_as :default
 
   def perform(days_to_check: 31)
-    users_with_recent_achievements = User.joins(:achievements).merge(Achievement.in_state(:complete))
-      .where("most_recent_achievement_transition.updated_at": days_to_check.days.ago..).distinct
-
-    missing_badges = []
-    users_with_recent_achievements.each do |user|
-      user.user_programme_enrolments.each do |user_programme_enrolment|
-        programme = user_programme_enrolment.programme
-        badge = programme.badges.active.first
-
-        next unless badge
-        next unless programme.badgeable?
-        next unless programme.user_qualifies_for_credly_badge?(user)
-        has_badge = user_has_badge?(user, programme)
-
-        if !has_badge
-          missing_badges << [user.email, programme.title]
-        end
-      end
-    end
-
     recent_achievements = Achievement.in_state(:complete).where("most_recent_achievement_transition.updated_at": days_to_check.days.ago..)
 
     missing_badges = []
@@ -33,11 +13,9 @@ class BadgeAssignmentCheck < ApplicationJob
         next unless badge
         next unless programme.badgeable?
         next unless programme.user_qualifies_for_credly_badge?(user)
-        has_badge = user_has_badge?(user, programme)
+        next if user_has_badge?(user, programme)
 
-        if !has_badge
-          missing_badges << [user.email, programme.title]
-        end
+        missing_badges << [user.email, programme.title]
       end
     end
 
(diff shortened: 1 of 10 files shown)


app/jobs/badge_assignment_check_job.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    missing_badges
+  end
+
+  private
+
+  def user_has_badge?(user, programme)
+    Credly::Badge.by_programme_badge_template_ids(user.id, programme.badges.pluck(:credly_badge_template_id))
+  end
```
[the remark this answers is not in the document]
msquance-stem: Yeah it is correct, as we would not allocate a badge to a user who had one of the previous ones, they do not get the active badge if they have an inactive one (which are the previous years badges)

app/jobs/badge_assignment_check_job.rb
```diff
@@ -0,0 +1,32 @@
+class BadgeAssignmentCheckJob < ApplicationJob
+  queue_as :default
+
+  def perform(days_to_check: 31, programmes_to_check: [])
+    missing_badges = []
+    programmes_to_check.each do |programme|
```
barrywoolgar, line 6: Minor code style note: If you were to use [`.each_with_object`](https://ruby-doc.org/3.1.4/Enumerable.html#method-i-each_with_object) here you wouldn't need to define `missing_badges`, or reference it at the end of the method. There's no functional difference though.

app/jobs/badge_assignment_check_job.rb
```diff
@@ -0,0 +1,32 @@
+class BadgeAssignmentCheckJob < ApplicationJob
+  queue_as :default
+
+  def perform(days_to_check: 31, programmes_to_check: [])
```
barrywoolgar, line 4: What do you think about making `programmes_to_check` a required argument?

barrywoolgar (approved):
Thanks Mike! I didn't notice before but (other than in the tests) this PR doesn't appear to _use_ the Job anywhere. Also, it's unusual for a Job to just return a value instead of taking some kind of action with it. If this job was queued with Sidekiq that return value would be lost, and the job would effectively become a no-op. I do see from your original comment that perhaps this job is for manual debugging purposes - if so I'd recommend you add a class comment detailing it's intended usage.

I've approved the PR as this all looks good to go, just a few minor suggestions/thoughts for if you're interested really. Have a great weekend!

msquance-stem:
Thanks for the pointers, I have made those few styling changes as suggested, and added some comments. The intention for this was just to be a debugging tool, I did consider that I might wrap this job into a rake task that then parses the results, but while working on it, I think I found the cause of the missing badges. So hopefully it won't be something we need in the future, but good to keep around for checking in the next few months to be sure.

Can you just give it one last pass then I can get it live?

barrywoolgar (approved):
Hi Mike, no problem at all, looks good to me!

PR #1915, opened 2024-01-29
Download option for lesson files in Unit view

msquance-stem:
## Status

* Review App link(s): https://teachcomputing-pr-1915.herokuapp.com/
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2571
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2584

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Adding download lesson button to the unit view. This will display the login button or the download button depending if the user is logged in or not.

Also fixing the typo I found in the header navigation

### DIFF
--- a/app/views/curriculum/units/_lessons.html.erb
+++ b/app/views/curriculum/units/_lessons.html.erb
@@ -3,7 +3,19 @@
 		<h3 class="govuk-heading-m">Lessons</h3>
 			<ul class="govuk-list govuk-body govuk-!-padding-bottom-7">
         <% lessons.each do |lesson| %>
-            <li class="curriculum__list--item <%= key_stage_list_color(key_stage.level) %>"><%= link_to lesson_title_wording(lesson), curriculum_key_stage_unit_lesson_path(key_stage_slug: key_stage.slug, unit_slug: unit.slug, lesson_slug: lesson.slug), class: 'govuk-link ncce-link' %></li>
+					<li class="curriculum__list--item curriculum__list--item-with-download <%= key_stage_list_color(key_stage.level) %>">
+						<%= link_to lesson_title_wording(lesson), curriculum_key_stage_unit_lesson_path(key_stage_slug: key_stage.slug, unit_slug: unit.slug, lesson_slug: lesson.slug), class: 'govuk-link ncce-link' %>
+
+						<% if current_user %>
+							<% if lesson.zipped_contents&.file.present? %>
+								<%= link_to generate_download_url(lesson.zipped_contents.file), class: 'govuk-button button button--aside lesson-download-button' do %>
+									Download all lesson files <span class="lesson-download-button__meta-data">(<%= lesson.zipped_contents.size %>)</span>
+								<% end %>
+							<% end %>
+						<% else %>
+							<%= link_to 'Log in to download', "#{auth_url}?source_uri=#{CGI.escape(request.original_url)}", method: :post, :class => 'govuk-button button button--aside' %>
+						<% end %>
+					</li>
         <% end %>
 			</ul>
 	</div>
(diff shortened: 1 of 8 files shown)


app/views/curriculum/units/_lessons.html.erb
```diff
@@ -3,7 +3,19 @@
 		<h3 class="govuk-heading-m">Lessons</h3>
 			<ul class="govuk-list govuk-body govuk-!-padding-bottom-7">
         <% lessons.each do |lesson| %>
-            <li class="curriculum__list--item <%= key_stage_list_color(key_stage.level) %>"><%= link_to lesson_title_wording(lesson), curriculum_key_stage_unit_lesson_path(key_stage_slug: key_stage.slug, unit_slug: unit.slug, lesson_slug: lesson.slug), class: 'govuk-link ncce-link' %></li>
+					<li class="curriculum__list--item curriculum__list--item-with-download <%= key_stage_list_color(key_stage.level) %>">
+						<%= link_to lesson_title_wording(lesson), curriculum_key_stage_unit_lesson_path(key_stage_slug: key_stage.slug, unit_slug: unit.slug, lesson_slug: lesson.slug), class: 'govuk-link ncce-link' %>
+
+						<% if current_user %>
+							<% if lesson.zipped_contents&.file.present? %>
```
markjs, line 10: Would we want to rearrange these conditionals slightly? This way round we could end up telling someone to log in and then once they log in there's nothing for them to download. Perhaps the `file.present?` check is around the whole block and then the logged in yes/no branches inside that?

spec/views/curriculum/units/_lessons_spec.rb
```diff
@@ -2,20 +2,41 @@
...
-    expect(rendered).to have_css(".govuk-list", count: 1)
+  context "user logged in" do
+    before do
+      allow_any_instance_of(AuthenticationHelper).to receive(:current_user).and_return(user)
+      render partial: "lessons", locals: {key_stage:, unit:, lessons: unit.lessons}
+    end
+
+    it "has download button" do
+      expect(rendered).to have_css(".govuk-button", text: "Download all lesson files")
+    end
```
markjs, line 40: We don't seem be testing the branch where there are no files to download. Would be good to make sure this test covers that case too unless we have a good reason not to.

markjs (changes requested):

app/views/curriculum/units/_lessons.html.erb
msquance-stem: Makes a lot of sense, will make those changes, thanks!

spec/views/curriculum/units/_lessons_spec.rb
msquance-stem: Totally agree, will add a test case for no files

markjs (approved):
Looks good to me and seems to work as expected. A couple of quick thougths:
- Could the logged out button say "Log in to download lesson files" to make it clearer exactly what can be downloaded and to be consistent with the logged in button.
- The downloads don't actually work but I'm expecting this is just a review app thing as they do go to an S3 URL which then itself 404s. Worth making sure the downloads definitely work once this goes live.
- The download button within a lesson which this PR hasn't changed has the conditionals the way round they were before my suggested change. This means when logged out you see a "log in to download" button and then you log in and the button disappears. Perhaps worth reordering those conditionals in `app/views/curriculum/lessons/show.html.erb` too if that's preferred behaviour.

Happy to mark this as "Approved" and you can decide if any of those above points are worth actioning or not.

msquance-stem:
Thanks Mark. 

- The programme team have already approved the text in the button so I think I will leave it as is for now
- Downloads don't work right on the review apps, but will check once live
- I have swapped the logic around on the show page, to match what we changed on the unit page. Added some testing for the non-file instances.

markjs (approved):

PR #1916, opened 2024-02-06
Fixing bug in auto enrolment for subject knowledge

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2587

## Review progress:

- [ ] Browser tested
- [x] Tech review completed

## What's changed?

Fixing bug due to nil variable

### DIFF
--- a/app/jobs/auto_enrol_job.rb
+++ b/app/jobs/auto_enrol_job.rb
@@ -20,7 +20,7 @@ class AutoEnrolJob < ApplicationJob
 
       Programmes::UserEnroller.new(
         programme_id: programme.id,
-        pathway_slug: pathway.slug,
+        pathway_slug: pathway&.slug,
         user_id: user.id,
         auto_enrolled: true
       ).call
--- a/spec/jobs/auto_enrol_job_spec.rb
+++ b/spec/jobs/auto_enrol_job_spec.rb
@@ -7,17 +7,19 @@ RSpec.describe AutoEnrolJob, type: :job do
 
       secondary_certificate = create(:secondary_certificate)
       primary_certificate = create(:primary_certificate)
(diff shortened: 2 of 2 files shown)


app/jobs/auto_enrol_job.rb
```diff
@@ -20,7 +20,7 @@ def perform(achievement:)
 
       Programmes::UserEnroller.new(
         programme_id: programme.id,
-        pathway_slug: pathway.slug,
+        pathway_slug: pathway&.slug,
```
markjs, line 23: This change looks pragmatic and the user enroller is already set up to handle a nil pathway_slug 👍

spec/jobs/auto_enrol_job_spec.rb
```diff
@@ -7,17 +7,19 @@
 
       secondary_certificate = create(:secondary_certificate)
       primary_certificate = create(:primary_certificate)
+      subject_knowledge = create(:cs_accelerator)
```
markjs, line 10: What is the reasoning for this variable being called `subject_knowledge` while the programme is also referred to as `cs_accelerator`? Are these names analogous? It seems inconsistent with the two lines above to use a different name but this isn't the only place that name is used like that so I don't have a strong feeling either way.

markjs (approved):
One question but not a deal-breaker 👍

spec/jobs/auto_enrol_job_spec.rb
msquance-stem: cs_accelerator has become subject_knowledge, and its been changed in some places in but not others, I started out calling it subject knowledge, then realised afterwards it was still called cs_accelerator in the factory and just forgot to update the variable

PR #1918, opened 2024-02-06
Seeding CP407

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2585

## Review progress:

- [ ] Tech review completed

## What's changed?

Seeding CP407

### DIFF
--- a/db/seeds/activities/stem_learning.rb
+++ b/db/seeds/activities/stem_learning.rb
@@ -2429,7 +2429,7 @@ Activity.find_or_initialize_by(stem_course_template_no: "fa051bb1-c5af-ee11-a569
 end.save!
 
 Activity.find_or_initialize_by(stem_course_template_no: "8f4f5edb-b3c1-ee11-9079-002248c6f979").tap do |activity|
-  activity.title = "Literacy via primary computing – building vocabulary and embedding literacy skills"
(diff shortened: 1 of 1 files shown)


abr-storm (dismissed):
A few things I noticed:
- We're using a unicode character instead of an ascii dash in the course title, could that cause problems?
- Do `category.remote_delivered_cpd = true` and `activity.category = "face-to-face"` contradict eachother?
- The ticket mentioned duration, but we're not encoding it anywhere - is that a detail we normally leave out?

Aside from that looks good

Edit: do you know why the deploy failed?

msquance-stem:
Thanks Alex, good spot on the hyphen, I have swapped that out.

The way we code a virtual course is to do a combination of "face-to-face" and remote_delivered_cpd. The face-to-face category is used a lot in the validation of course completion, but now many of the face to face courses are done remote. Which is different to the online courses, so this is the way they have been coded.

Duration is how we calculate credits, in this instance, Dapo has given us the credits, to save us the maths. The credits are duration divided by 6. But there is no where this is coded, it eventually gets pulled over from Dynamics (through the achiever services) so we don't store it.

abr-storm (approved):
LGTM

PR #1919, opened 2024-02-06
Changes to navigation

msquance-stem:
## Status

* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2588

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Removing school from primary and secondary
Removing the bold formatting on first item
Tweaked the min size to give a more satisfying layout

### DIFF
--- a/app/webpacker/stylesheets/components/_dropdown-expander.scss
+++ b/app/webpacker/stylesheets/components/_dropdown-expander.scss
@@ -16,6 +16,7 @@
   padding: 1.125rem 1rem 1rem 1rem;
   visibility: hidden;
   width: stretch;
+  min-width: 10rem;
 
   @include govuk-media-query($from: desktop) {
     border-bottom: unset;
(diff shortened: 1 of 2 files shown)


markjs:
Functionally this all works fine, but I wonder if the spacing on the menu could look a bit tidier.

Screenshot 2024-02-07 at 11 47 51

The extra space here after "Primary" just looks a bit awkward to me but I understand it is there to ensure the dropdown isn't too narrow.

Screenshot 2024-02-07 at 11 47 59

However I think even with the extra spacing that dropdown is awkwardly narrow. "Early career and trainee teachers" spanning across three lines like that looks a bit pants. Could we restructure the markup so the width of the dropdown doesn't need to be the same width as the top level button and then we can adjust the spacing better? What do you think @msquance-stem? Perhaps one for a designer to cast an eye over and suggest the best spacing.

Just fiddling around in browser with CSS properties I think this looks a lot tidier with reduced spacing on the top level button and a wider dropdown:
Screenshot 2024-02-07 at 11 56 15

markjs (approved):

PR #1920, opened 2024-02-06
Add anchor links to each year of a key stage

abr-storm:
This will enable users to share links to a specific year.

We may want to bear in mind this GOVUK guidance that suggests against using anchor links:
> Try to avoid using anchor links in your content. Anchor links can be disorientating for some users with access needs who may have problems getting back to the previous page.
> 
> For example, screen magnifier users and those with motor function impairments. People using assistive technology may find it hard to see what they’ve missed or realise that they’ve skipped forward.
> 
> Avoid using anchor links to content on the same page. Instead, rearrange the structure of your content so that people can navigate the content more easily.

Instead of making the headings themselves the link, I've added small links beneath the headings so we can use the normal link styles:
image

### DIFF
--- a/app/helpers/curriculum_helper.rb
+++ b/app/helpers/curriculum_helper.rb
@@ -20,6 +20,10 @@ module CurriculumHelper
     "Year #{year_number}"
   end
 
+  def year_group_anchor(year_number)
+    year_group_title(year_number).parameterize
+  end
+
   def user_has_rated?(id)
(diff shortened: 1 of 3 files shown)


abr-storm:
closes #1920

msquance-stem (approved):
LGTM

hannah-storm (approved):

PR #1924, opened 2024-02-08
Feature/new non gcse page

abr-storm:
closes https://github.com/NCCE/teachcomputing.org-issues/issues/2593


markjs:
Screenshot 2024-02-19 at 16 24 46

Is this (left hand button in the screenshot) the correct hover state for these buttons? The filled in green buttons go entirely black on hover, border included, which I think would look better here. If this is the desired style or a style used elsewhere then that's fine, but that's not clear from the figma file.

markjs (approved):

PR #1925, opened 2024-02-09
Update to senior leaders pages

msquance-stem:
## Status

* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2540

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Two pages rebuilt
Creation of a new couple of shared banner elements


abr-storm (dismissed):
Here are some notes I made on the two pages. There a few little things, but as far as I'm concerned the only showstopper is the "Contact you local computing hub" typo

<details>
  <summary>Secondary Senior Leaders</summary>

---

The body text here is grey, but in the design it's black

image


---
The card is offset wrt. the text, whereas in the design it's flush

image

--- 

The headings in this aside look like they have a higher font weight than the design

The "Contact you local computing hub" button isn't in the design

image

---

The body text here is black, whereas in the design it looks slightly grey

image

---

In the design the word "subsidies" is a link

The font weight of the main link at the bottom looks higher than the design (applies to sibling cards too)

image

---

The design has an external link icon here, but given that the GOVUK advice is not to use those, maybe we should leave it out?

image


---

In the design the "Teaching secondary computing" link is underlined

Same comment about the main link font weight (and siblings)

image

</details>

<details>
  <summary>Primary Leaders Page</summary>

---

Has a lot of the same stuff as the secondary one in places where elements are shared

---

The "Register now" button is wider on the design
image

</details>

abr-storm (dismissed):
I noticed a couple further things, but I'm aware that the design has been changing as we go - so I think we're close enough after we correct that new button typo.

Body text is still grey here, whereas it's black in the design:
image

"Find our more" typo on this button:
image

Link is underlined here, it isn't on the design:
image

Body text is black here, whereas it's grey in the design:
image

app/views/pages/primary-senior-leaders.html.erb
```diff
@@ -1,187 +1,152 @@
...
-              ),
-              link: {
-                link_title: t('.how-we-help.cards.courses.link.title'),
-                link_url: t('.how-we-help.cards.courses.link.url'),
-                tracking_page: 'Primary SLT',
-                tracking_label: t('.how-we-help.cards.courses.link.event-label')
-              }
+              title: t('.objectives.cards.confidence.title'),
+              text: t('.objectives.cards.confidence.text.html',
+                  subsidies_link: (link_to('subsidies', cms_page_url('funding#primary-subsidies')))),
```
barrywoolgar, line 73: The URL appears to be broken, I think it should be:

`cms_page_url('funding', anchor: 'primary-subsidy')` 

Note the singular spelling, and no need for hash symbol in anchor value (some that needs to be fixed elsewhere in the project).

app/views/pages/secondary-senior-leaders.html.erb
```diff
@@ -1,227 +1,161 @@
...
-              ),
-              link: {
-                link_title: t('.how-we-help.cards.courses.link.title'),
-                link_url: t('.how-we-help.cards.courses.link.url'),
-                tracking_page: 'Secondary SLT',
-                tracking_label: t('.how-we-help.cards.courses.link.event-label')
-              }
+              title: t('.objectives.cards.invest.title'),
+              text: t('.objectives.cards.invest.text.html',
+                  subsidies_link: (link_to('subsidies', cms_page_url('funding#secondary-subsidies')))),
```
barrywoolgar, line 81: This link is also broken for the same reason

config/locales/views/pages/secondary-senior-leaders/en.yml
```diff
@@ -2,107 +2,54 @@ en:
...
             link:
-              title: "Start or join a club"
-              url: "https://codeclub.org/en/"
-              event-label: "Code Club"
-          bebras:
-            title: "Bebras UK"
-            text: "The Bebras Computing Challenge introduces computational thinking to students. It is organized in over 50 countries and designed to get students all over the world excited about computing."
+              title: "Discover Isaac Computer Science"
+              url: "/issac_computer_science"
+              event-label: "issac computer science"
```
barrywoolgar, line 47: This URL looks like it should be `/isaac-computer-science` instead. Also beware the misspelling of "issac" here.

barrywoolgar (changes requested):
There are some broken links that need fixing here, and a number of design deviations. I'm not sure how strict we need to be with the latter, so please let me know if these are acceptable and I'll approve (after the links are fixed).

Design deviations apply to both pages:

The font-size and line-spacing of the top section (on green background) appears quite different from the design. Is this a global issue?

The first paragraph of "Computing Quality Framework" differs from the design.

The spacing between "Computing Quality Framework" and "Guidance for offering computing" appears to be half that of the design.

Margin beneath "Computing Quality Framework" box appears to be double that of design.

The font weight of the links at the bottom of the three boxes seems higher than the design.

"Brochure" in "Read our Brochure" is not capitalised on design, and appears out of keeping with existing capitalisation styles.

The three boxes under "Resources" have a drop shadow not present on the design. They also appear to have slightly more internal padding which might allow the text to flow better if it was reduced (but perhaps madness this way lies). There also appears to be a different amount of margin above/below these boxes compared to the design.

Additionally, the three "Resources" boxes appear to have their final links bottom-aligned, which is inconsistent with the earlier boxes and the design.

Thanks

abr-storm:
Ok, the links and copy should be fixed now. The styling discrepancies appear to be global issues between the components being used in the app and the designs, which are out of scope for this PR.

I've also reverted the earlier changes I made to the text color after talking to the design team.

Please could you have another look when you're ready @barrywoolgar?

barrywoolgar (dismissed):
Looks good to me, thanks Alex

Caleb-T-Owens (approved):
LGTM 👍

PR #1926, opened 2024-02-12
Remove links from key stage pages, but still allow external linking

abr-storm:
Resolves https://github.com/NCCE/teachcomputing.org-issues/issues/2592#issuecomment-1936037350

This PR removes anchor links from the key stage pages, but retains the generated IDs so the sections can be linked to externally.

It also adds these links to the sitemap


msquance-stem (approved):
LGTM

PR #1938, opened 2024-02-13
2597/strapi service

msquance-stem:
## Status

* Current Status: WIP
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2597

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

First pass at an abstracted CMS service


app/components/benchmark_banner_component/benchmark_banner_component.scss
```diff
@@ -0,0 +1,58 @@
...
+
+.benchmark-banner--width-float {
+  width: 100%;
+
+  .benchmark-banner {
+    margin-top: 10px;
+  }
+
+  @include govuk-media-query($from: desktop) {
+    width: calc(((100vw - 960px)/2) + (960px * 0.333) - 30px);
```
Caleb-T-Owens, line 41: There are a lot of magical constants here. It would be really nice if we could have the width determined by the parent element.

If not, it would be great if this maths could be simplified. I believe the following is what is evaluates to:
```
width: calc(50vw - 190px);
```

app/components/benchmark_bordered_cards_component.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class BenchmarkBorderedCardsComponent < ViewComponent::Base
+  def initialize(cards:, class_name: nil, cards_per_row: 2)
+    @cards = cards
+    @class_name = class_name
+    @cards_per_row = cards_per_row
+  end
+
+  def custom_properties
+    "--cards-per-row: #{@cards_per_row};"
```
Caleb-T-Owens, line 11: Using a CSS var like this is very nice!

app/components/bordered_cards_component.rb
```diff
@@ -2,7 +2,16 @@
...
+    # initially this component only supported one link per card, but now we need
+    # to add mutiple links per card. old callers may be using cards with a :link
+    # field, so we copy that over to a :links field with one one element
+    @cards = cards.map! do |card|
+      card[:links] ||= []
+      card[:links] << card[:link] if card.key?(:link)
+      card.delete(:link)
+      card
+    end
+
```
Caleb-T-Owens, line 14: Could we decide whether we want to be a map or an each please 😆 

```suggestion
    @cards = cards

    @cards.each do |card|
      card[:links] ||= []
      card[:links] << card[:link] if card.key?(:link)
      card.delete(:link)
    end
```

app/components/course_component/course_component.html.erb
```diff
@@ -12,7 +12,7 @@
     </div>
   <% elsif @course.occurrences.present? %>
     <div class="govuk-body-s govuk-!-font-weight-bold ncce-courses__locations">
-      <button aria-expanded="false">
+      <button type="button" aria-expanded="false">
```
Caleb-T-Owens, line 15: Why do we need this annotation here?

PR #1956, opened 2024-02-21
Fixing mixup in legacy course definition

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2606

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Fixing mix up with course changes from the other week


msquance-stem:
Yeah, its messy, but there are two badges, the non-secondary one should not be in here, so has been marked as legacy just in case anyone has it linked, and the secondary one is legacy false on line 82

bhagdave (approved):

PR #1968, opened 2024-03-05
Improved UPE pending time logic and movement of certificate number generation

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2613

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- To allow programme team to complete users, we have moved the certificate number generation to the state machine, out of the timed job
- Due to the ability to rollback, and this possibly occurring multiple times within the "pending" period, the flag on User Programme Enrolments is not sufficient to prevent accident completion
- Rather than just auto move them to completed at the end of the timer. The timer now checks if the UPE is in the pending state, if so checks if the last change occurred equal or longer than the pending period ago. If so will transition, if not will do nothing
- Updated the job to allow for instant to movement of UPE to completed when no pending period required
- Required changes to test suite to ensure programmes had counters setup


app/models/state_machines/user_programme_enrolment_state_machine.rb
```diff
@@ -21,8 +21,12 @@ class StateMachines::UserProgrammeEnrolmentStateMachine
     enrolment.programme.mailer.with(user: enrolment.user).pending.deliver_later
   end
 
-  after_transition(to: :complete) do |programme_enrolment|
+  before_transition(to: :complete) do |programme_enrolment, transition|
     # Set the name that should display on the certificate
+    transition.metadata["certificate_name"] = programme_enrolment.programme.programme_complete_counter.get_next_number
```
abr-storm, line 26: Is this meant to be `certificate_number`? It looks like we immediately override the `certificate_name` in `set_user_programme_enrolment_complete_data` and then try to pull out a `certificate_number` from the transition metadata in the certificate generator.

```suggestion
    transition.metadata["certificate_number"] = programme_enrolment.programme.programme_complete_counter.get_next_number
```

app/jobs/schedule_certificate_completion_job.rb
```diff
@@ -2,7 +2,10 @@ class ScheduleCertificateCompletionJob < ApplicationJob
   queue_as :default
 
   def perform(enrolment, meta = {})
-    certificate_number = enrolment.programme.programme_complete_counter.get_next_number
-    enrolment.transition_to(:complete, meta.merge({certificate_number: certificate_number}))
+    return unless enrolment.in_state?(:pending)
+
+    if enrolment.last_transition.created_at <= (DateTime.now - enrolment.programme.pending_delay)
+      enrolment.transition_to(:complete, meta)
+    end
```
abr-storm, line 9: Like you suggested, maybe we could add some margin here to account for jobs being delayed?

```suggestion
    return unless enrolment.in_state?(:pending)
    return unless enrolment.last_transition.created_at <= (enrolment.programme.pending_delay.days.ago + 30.minutes)

    enrolment.transition_to(:complete, meta)
```

app/models/state_machines/user_programme_enrolment_state_machine.rb
msquance-stem: You are correct, I have used the wrong one!

app/models/state_machines/user_programme_enrolment_state_machine.rb
```diff
@@ -21,8 +21,14 @@ class StateMachines::UserProgrammeEnrolmentStateMachine
...
-  after_transition(to: :complete) do |programme_enrolment|
-    # Set the name that should display on the certificate
+  before_transition(to: :complete) do |programme_enrolment, transition|
+    next unless transition.metadata["certificate_number"].blank?
+
+    # Set the number that should display on the certificate
+    transition.metadata["certificate_number"] = programme_enrolment.programme.programme_complete_counter.get_next_number
+  end
+
+  after_transition(to: :complete) do |programme_enrolment, transition|
```
Caleb-T-Owens, line 31: The transition param is unused here, could we please remove it

Caleb-T-Owens (approved):
One small comment that can be ignored if you this wants to just go live 👍

Caleb-T-Owens (approved):

PR #1970, opened 2024-03-06
Revert "Bump cloudflare-rails from 1.2.0 to 5.0.1"

abr-storm:
It looks like we rely on some undocumented behavior that only exists < 2.0 to get a list of cloudflare Ips.

This causes our rack attack config to fail in staging.

This reverts commit 6d5f4c3c4757400b33688dc162d835d93d5f29e6.


Caleb-T-Owens (approved):
LGTM 👍

PR #1972, opened 2024-03-06
New school trusts page

abr-storm:
Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2614

image
image


vickymg:
I've just checked this out on staging, and mostly looks good! One really minor thing, but could we just put the testimonial text inside quote marks as per the design? Just makes it clearer that it's a quote rather than copy!

PR #1986, opened 2024-03-12
Updating setter to downcase Stem Course Template No

msquance-stem:
Adding test cases and updating seed
Updates need to be done in the console before deploying this, to prevent courses being double seeded

## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2628

## Review progress:

- [ ] Tech review completed

## What's changed?

- Making the setter automatically downcase the ID
- Updated the seed file

## Steps to perform before deploying to production

The ID's listed in the issue need to be downcased **before** deployment to production otherwise the seed will generate a second copy of the course.

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -23,7 +23,7 @@ class Activity < ApplicationRecord
   }
   validates :future_learn_course_uuid, uniqueness: true, unless: proc { |a| a.future_learn_course_uuid.blank? }
   validates :stem_activity_code, uniqueness: true, unless: proc { |a| a.stem_activity_code.blank? }
-  validates :stem_course_template_no, uniqueness: true, unless: proc { |a| a.stem_course_template_no.blank? }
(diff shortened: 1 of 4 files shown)


app/models/activity.rb
```diff
@@ -53,6 +53,11 @@ def self.cs_accelerator_diagnostic_tool
     end
   end
 
+  def stem_course_template_no= value
+    # Ensure that all stem_course_template_no's are stored in downcase format
+    super(value.downcase)
+  end
+
```
abr-storm, line 60: It's a shame we're not on 7.1 yet, we could have used the new normalize feature here https://www.bigbinary.com/blog/rails-7-1-adds-activerecord-base-normalizes

abr-storm (approved):

app/models/activity.rb
Caleb-T-Owens: @msquance-stem could you please use brackets around your setter's arguments?

`def stem_course_template_no=(value)`

Caleb-T-Owens (changes requested):
One style comment

Caleb-T-Owens (approved):
LGTM

PR #1990, opened 2024-03-13
Fix/i belong video breakpoint

abr-storm:


Caleb-T-Owens (approved):
I've taken a look on multiple screen sizes and it looks good to me 👍

PR #1993, opened 2024-03-14
Update chrome version

abr-storm:


dcorking:
@Caleb-T-Owens it might be worth a quick check to see if you still need the hardcoded chrome version you added in line 82 https://github.com/NCCE/teachcomputing.org/commit/60796329fe5cd3a86a9e91479ca1d751ee4793b5

Caleb-T-Owens:
@dcorking I'll have a go, but in the past it was definitely required because our selenium version couldn't work with the latest chrome

Caleb-T-Owens (approved):
LGTM 👍

dcorking:
Good to know: our future selves will probably thank us for a code comment reminding them to sync the Chrome version with Selenium. (Main thing is that I'm glad it works as Chrome got in a bad habit of deleting old versions of the installer a year or two ago)

PR #1998, opened 2024-03-18
Homepage update

abr-storm:
## Status

* Current Status: With STEM to review

* Related to https://github.com/NCCE/teachcomputing.org-issues/issues/2621

## Review progress:

- [x] Browser tested
- [x] Tech review completed
- [x] STEM review completed


spec/system/pages/home_spec.rb
```diff
@@ -8,15 +8,25 @@
...
+      ".axe-skip-a11y-test"
+      "ncce-home-hero__house-background-corner"
+      "ncce-home-hero__house-roof"
+      "ncce-home-hero__house-wall-left"
+      "ncce-home-hero__house-wall-right"
+      "ncce-home-hero__house-base"
+      "ncce-home-hero__person-corner-left"
+      "ncce-home-hero__person-corner-right"
+    ]
+    expect(page).to be_accessible.within("#main-content").excluding(exclusions)
```
Caleb-T-Owens, line 29: Could we instead make the page accessible rather than just ignoring it 😆?

For each of those images, if you just add an area-hidden, it ought to do the trick

Caleb-T-Owens (changes requested):
Looks pretty good, just some accessability tweaks that would be good to make

spec/system/pages/home_spec.rb
croab: So I had added the aria_hidden fields but will look into why this isn't preventing the tests from failing 👀

Caleb-T-Owens (approved):
Thanks for all the work on this! End result is looking really nice

app/views/pages/home/_confidence.html.erb
```diff
@@ -1,51 +1,51 @@
 <div class="govuk-width-container">
-  <div class="ncce-confidence__header govuk-heading-m">
+  <div class="ncce-confidence__header govuk-heading-s">
     <h1><%= t("pages.home.confidence.heading") %></h1>
   </div>
 
   <div class="govuk-main-wrapper ncce-confidence">
     <div class="ncce-confidence__card">
-      <%= image_pack_tag 'media/images/home/card_courses.jpg', loading: 'lazy', class: 'ncce-confidence__icon', alt: '' %>
+      <%= link_to image_pack_tag('media/images/home/card_courses.jpg', loading: 'lazy', class: 'ncce-confidence__icon', alt: ''), courses_path %>
```
msquance-stem, line 8: For these links, if you move the class onto the link, rather than the image. The original justify-self method will work for the alignment on mobile view

app/views/pages/home/_confidence.html.erb
```diff
@@ -1,51 +1,51 @@
...
+      <%= link_to t("pages.home.confidence.teaching_resources.title"), curriculum_key_stages_path, class: "ncce-confidence__link" %>
       <p class="ncce-confidence__description">
         <%= t("pages.home.confidence.teaching_resources.description") %>
       </p>
     </div>
 
     <div class="ncce-confidence__card ncce-confidence__card--orange">
-      <%= image_pack_tag 'media/images/home/card_i_belong.jpg', loading: 'lazy', class: 'ncce-confidence__icon', alt: '' %>
-      <%= link_to t("pages.home.confidence.i_belong.title"), i_belong_path, class: "ncce-confidence__link" %>
+      <%= link_to image_pack_tag('media/images/home/card_i_belong.jpg', loading: 'lazy', class: 'ncce-confidence__icon', alt: ''), "/i-belong" %>
```
msquance-stem, line 32: Can you swap the hard coded i-belong link to use the `about_i_belong_path` helper method. We should use the path helpers where possible rather than hard coded urls.

app/webpacker/stylesheets/components/pages/home/_confidence.scss
```diff
@@ -75,8 +75,8 @@
         grid-column: 2;
         width: 87px;
         height: 53px;
-        justify-self: end;
         border: none;
+        float: right;
```
msquance-stem, line 79: If you make the class changes suggested in the other file, we can use the `justify-self: end;` rather than the float.

msquance-stem (changes requested):

app/views/pages/home/_confidence.html.erb
A-Wheeto: Moved the class onto all the links

app/webpacker/stylesheets/components/pages/home/_confidence.scss
A-Wheeto: Reverted back to justify-self after other changes.

msquance-stem (approved):
Thanks for those changes. LGTM

PR #2007, opened 2024-03-28
2641/add virtual storage career video

msquance-stem:
## Status

* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2641

## Review progress:

- [x] Browser tested
- [x] Tech review completed

## What's changed?

- Created a new banner component
- Added in the video to the required sections


app/components/banner_component.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  def color_class color
+    "#{color}-bg" if color
+  end
+
+  def text_column_class
+    classes = ["banner-component__content-section"]
+    classes << case @media_column_size
+    when :third
+      "banner-component__content-section--two-thirds"
```
markjs, line 53: Is this value supposed to be the same as the one in the `else` clause? If so then it can of course be simplified to a single `if/else`, otherwise this class may need changing

app/components/banner_component/banner_component.html.erb
```diff
@@ -0,0 +1,36 @@
...
+          <div class="banner-component_title">
+            <h2 class="govuk-heading-m">
+              <%= @title %>
+            </h2>
+          </div>
+          <div class="banner-component_body">
+            <p class="govuk-body">
+              <%= @text %>
+            </p>
+            <% if @link %>
```
markjs, line 16: If the "empty" state of `@link` is just `nil` then this if statement isn't really achieving anything. What do you think to getting rid of it and just having `<%= @link %> in isolation? Alternatively it could be changed to `if @link.present?` or something like that?

Or does the `<%= @link %>` part need to be smarter and output the text within a `link_to` tag?

app/components/banner_component.rb
```diff
@@ -0,0 +1,61 @@
...
+      "banner-component__media-section--one-third"
+    when :half
+      "banner-component__media-section--half"
+    else
+      "banner-component__media-section--one-third"
+    end
+    classes.join(" ")
+  end
+
+  def color_class color
```
markjs, line 45: Is standardrb happy with leaving these parentheses out? Looks gross to me but that might just be me 😂

spec/components/banner_component_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+    before do
+      render_inline(described_class.new(
+        title: "Banner title",
+        text: "Some random text",
+        image: {
+          url: "media/images/test/image.png",
+          alt: ""
+        },
+        background_color: "purple",
+        link: "http://www.example.com"
```
markjs, line 14: No tests here to actually assert that the title, text, and link are being output. Seems worth having a quick one to check the values passed in are being respected.

markjs (approved):
A few questions / suggested changes. I'm happy for you to merge if you don't think any further action is needed based on my comments or for you to action them and then merge afterwards. Cheers

app/components/banner_component/banner_component.html.erb
msquance-stem: Thanks, this was an old bit of code that is no longer used, but I think it might be needed in the future. Have modified this take in an object instead, and now uses link_to to produce a better result

app/components/banner_component.rb
msquance-stem: I had in my head there might be more options, but I think thats not the case, so have simplified it to an IF statement

markjs:
Latest changes look great @msquance-stem :shipit:

PR #2009, opened 2024-04-03
Update tcc maps

croab:
## Status

* Current Status: WIP (pending tests)

## Review progress:

- [x] Browser tested
- [ ] Tech review completed

## What's changed?

New curriculum journey maps have been added previously to the S3 bucket. Changes made by myself here have been to create the UI for these maps which has involved:
- Copy changes to text for the current full curriculum journey (and addition of text/links for primary and secondary journeys to this page)
image

- Addition of icons/text/links for curriculum journeys to primary and secondary to their own key stage show pages (respectively)
image

- Test changes for these primary and secondary curriculum journeys
- New icons for the full, primary and secondary journeys

## Outstanding issues

* Awaiting confirmation of required GTM labels from @msquance-stem  

## Steps to perform prior deploying to production

* Files on the S3 bucket and their paths in the app should be updated such that the existing full curriculum journey path does not change (as requested by Dapo).


Caleb-T-Owens:
We have discussed some changes in person that are going to be made to this PR

spec/views/curriculum/key_stages/show_spec.rb
```diff
@@ -23,6 +23,14 @@
       expect(rendered).to have_css(".hero__heading", text: "Key stage 1 resources")
     end
 
+    it "links to the journey poster" do
+      key_stage = JSON.parse(unit_json, object_class: OpenStruct).data.key_stage.level
+      key_stage_type = ["1", "2"].include?(key_stage) ? "Primary" : "Secondary"
+      key_stage_range = (key_stage_type == "Primary") ? "1-2" : "3-4"
+
+      expect(rendered).to have_link("viewing and progressing through our #{key_stage_type} curriculum journey", href: "https://static.teachcomputing.org/journey/Curriculum.Journey_#{key_stage_type}.#{key_stage_range}.pdf")
+    end
```
Caleb-T-Owens, line 32: I'm generally not a fan of very complicated logic in tests like this, because we've essentially got a mirror of what the real code should be doing.

IMO tests shouldn't have any conditionals in them. They should construct the exact data required and then assert that the subject method/class/page does what you expect it to when given the particular data you've constructed and passed to it.

Could this be simplified so the link in that "have_link" assertion has the correct values just being part of the string?

app/helpers/curriculum_helper.rb
```diff
@@ -14,6 +14,21 @@ def generate_download_url(contents_url)
...
+    stage = if primary_years.include?(key_stage)
+      "primary"
+    elsif secondary_years.include?(key_stage)
+      "secondary"
+    end
+
+    return stage.upcase_first if upcase_first
+
+    stage
+  end
```
Caleb-T-Owens, line 30: Is this still used?

app/components/aside_component.rb
```diff
@@ -23,4 +24,10 @@ def tracking_data(label = nil)
       event_label: label
     }
   end
+
+  def external_image?(image)
+    return if image[:file].blank?
+
+    image[:file].start_with?('http://', 'https://')
+  end
```
Caleb-T-Owens, line 32:
```suggestion
def aside_image_tag(image, *args, **)
  if image[:file].start_with?('http://', 'https://')
    image_tag(image, *args, **)
  else
    image_pack_tag(image, *args, **)
  end
end

File: app/components/aside_component/aside_component.html.erb [hunk 4]
```diff
@@ -1,7 +1,7 @@
 <div class="aside-component<%= " #{@class_name || ''}" %>">
   <% if @image.present? %>
     <div class="aside-component__image">
-      <%= link_to image_pack_tag(@image[:file], alt: @image[:alt], title: @image[:title]),
+      <%= link_to send(@external_image ? :image_tag : :image_pack_tag, @image[:file], alt: @image[:alt], title: @image[:title]),
```
Caleb-T-Owens, line 4:
```suggestion
<%= link_to aside_image_tag(@image[:file], alt: @image[:alt], title: @image[:title]),

Caleb-T-Owens (approved):
LGTM 👍

PR #2010, opened 2024-04-04
Add new/improved icon

croab:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2631

## Review progress:

- [x] Browser tested
- [x] Tech review completed

## What's changed?

Set the value of the bool `is_new_course` on the `@course` object passed to the course component. This has been currently set to return `true` if `activity.created_at > 3.months.ago`

N/A

### DIFF
--- a/app/components/course_component/course_component.html.erb
+++ b/app/components/course_component/course_component.html.erb
@@ -1,4 +1,9 @@
 <div class="ncce-courses__course">
+  <% if @course.new_course? %>
+    <p class="govuk-body ncce-courses__new-improved-badge">
+      <%= "New/improved" %>
+    </p>
+  <% end %>
   <h2 id="<%= @course.title %>" class="govuk-heading-s ncce-courses__heading">
     <%= link_to @course.title, course_path(id: @course.activity_code, name: @course.title.parameterize), class: 'ncce-link' %><span class="ncce-courses__heading-code"><%= @course.activity_code %></span>
   </h2>
--- a/app/services/achiever/course/template.rb
+++ b/app/services/achiever/course/template.rb
@@ -22,7 +22,8 @@ class Achiever::Course::Template
     :topics_covered,
     :who_is_it_for,
     :workstream,
-    :always_on
+    :always_on,
+    :last_updated_at
 
   RESOURCE_PATH = "Get?cmd=CourseTemplatesListingByProgramme".freeze
   QUERY_STRINGS = {Page: "1",
@@ -66,6 +67,7 @@ class Achiever::Course::Template
       t.workstream = resource["Template.Workstream"]
 
       t.always_on = activity&.always_on || false
+      t.last_updated_at = activity&.created_at || Date.current
     end
   end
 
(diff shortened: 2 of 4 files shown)


app/services/achiever/course/template.rb
```diff
@@ -66,6 +67,7 @@ def self.from_resource(resource, activity)
       t.workstream = resource["Template.Workstream"]
 
       t.always_on = activity&.always_on || false
+      t.is_new_course = activity.created_at > 3.months.ago
```
Caleb-T-Owens, line 70: The ticket is wanting to have this tag displayed whenever a CPD is created or reauthored.

This looks like its using the `created_at` timestamp from the activity record in our database. Does this mean that the tag will only show three months after its been created in our database? I'm wondering if there is some `Template.CreatedAt` or similar timestamp that gets provided from Dynamics that we could make use of instead. I'm also wondering if a `created_at` timestamp is the correct thing to use, or whether we should instead be making use of an `updated_at` or even an `authored_at` like timestamp... I thing that would probably depend on what reauthoring looks like on the dynamics side though.

app/components/course_component/course_component.html.erb
```diff
@@ -1,4 +1,7 @@
 <div class="ncce-courses__course">
+  <% if @course.is_new_course %>
+    <%= image_pack_tag('media/images/icons/new-improved-icon.svg', class: "ncce-courses__badge", alt: '') %>
```
Caleb-T-Owens, line 3: It would be nicer to have this tag displayed using html & css rather than using a SVG image, so its easier to change in the future, similar to the "this term" tag on the enrichment page, or the "featured" tag on the home page articles


If you stick with SVG image, could you add an alt tag here so screen readers can correctly display them, or if we don't want them to show, instead use aria tags to make it hidden.

app/services/achiever/course/template.rb
```diff
@@ -22,7 +22,8 @@ class Achiever::Course::Template
     :topics_covered,
     :who_is_it_for,
     :workstream,
-    :always_on
+    :always_on,
+    :is_new_course
```
Caleb-T-Owens, line 26: Would it be nicer to have a method called `new_course?` that returns the value, rather than `is_new_course`?

spec/components/course_component_spec.rb
```diff
@@ -29,6 +29,20 @@
...
   end
 
+  context "new/improve course icons" do
+    it "shows a new/improved icon on new courses" do
+      course.is_new_course = true
+      render_inline(described_class.new(course: course, filter: filter))
+      expect(page).to have_css("img.ncce-courses__badge")
+    end
+
+    it "show not show a new/improved icon on old courses" do
```
Caleb-T-Owens, line 39:
```suggestion
    it "should not show a new/improved icon on old courses" do
```

Caleb-T-Owens (changes requested):
It would be nice to reconsider the usage of the SVG image, and to double check that we're implementing the expected behaviour for when the tag should be displayed.

Thanks 👍

Caleb-T-Owens (approved):
All looks good code wise 👍

I'm approving this, though let's wait to hear back from vicky before merging in case we need to change the condition

PR #2011, opened 2024-04-04
Banner for subject knowledge certificate

croab:
## Status

* Current Status: WIP
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2647

## Review progress:

- [x] Browser tested (with some conditions set in console)
- [x] Tech review completed

## What's changed?

Added an Alerts component for the heading which (for the current task) shows a tooltip with a link to the incomplete Subject Knowledge programme.

## Outstanding issues

The query to get the relevant incomplete enrolments to programmes uses the `updated_at` value on the user enrolment as a means of showing the notification only for 48 hours. Ideally would use transitions data but it doesn't look like this is persisted. Will have another look in the morning.

No tracking has been added to the link in the alert as there appears to be precedent (e.g. completed_face_to_face_course.html.erb doesn't appear to use tracking when letting enrolled users return to the subject knowledge certificate).

N/A


app/components/notification_component/notification_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<div class="notification-component <%= @wrapper_css %>">
+  <div class="tooltip">
+    <%= image_pack_tag('media/images/icons/alert.svg') %>
+    <div class="tooltiptext">
+      <% @notifiable_programmes.each do |programme| %>
+        <p class="govuk-body-s">
+          <%= t('.text') %>
+        </p>
+        <%= link_to t('.link_text'), notification_link(programme), class: 'govuk-link' %>
```
Caleb-T-Owens, line 9: rather than having a helper, could there be a #notification_link method on the programme models?

app/models/programme.rb
```diff
@@ -60,6 +60,17 @@ def enough_activities_for_test?(_user)
...
+  def show_notification_for_test?(user)
+    false
+  end
+
+  def in_notifiable_period?(user)
+    enrolment = user.user_programme_enrolments.find_by(programme_id: id)
+    notifiable_date_range = 48.hours.ago..Time.current
+
+    notifiable_date_range.cover?(enrolment.last_enrolment_date)
+  end
```
Caleb-T-Owens, line 72: This method feels like an implementation detail that is only relevant for CSA. Could this be moved onto the CSA model?

app/models/user_programme_enrolment.rb
```diff
@@ -31,6 +31,12 @@ def completed_at?
     last_transition.created_at
   end
 
+  def last_enrolment_date
+    return last_transition.created_at if state_machine.history.any?
+
+    created_at
+  end
```
Caleb-T-Owens, line 38: Could this be called `last_enrolled_at`?

Caleb-T-Owens (changes requested):
A few minor tweaks, but on the whole looking really good

Caleb-T-Owens (approved):
lgtm

PR #2034, opened 2024-05-10
Improve test coverage and fixing i-belong inactivity queries

msquance-stem:
Improving test coverage on the jobs and some of the rake tasks
Discovered that the i_belong inactivity queries were not working correctly so have fixed those

## Status

* Current Status: Ready for tech
* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2668

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Improved test coverage on jobs listed in the ticket
- Added some test cases to increase coverage on tasks
- Found an issue in testing around the i-belong inactivity queries, these should now be working correctly


spec/jobs/cs_accelerator/check_next_steps_job_spec.rb
```diff
@@ -1,12 +1,23 @@
...
+  let(:non_csa_activity) { create(:activity) }
+  let(:programme_activity) { create(:programme_activity, activity:, programme: cs_accelerator) }
+  let(:programme_activity_2) { create(:programme_activity, activity: activity_2, programme: cs_accelerator) }
+  let(:non_csa_programme_activity) { create(:programme_activity, activity: non_csa_activity, programme: other_programme) }
+  let(:achievement) { create(:achievement, user:, activity:) }
+  let(:achievement_2) { create(:achievement, user:, activity: activity_2) }
+  let(:non_csa_achievement) { create(:achievement, user:, activity: non_csa_activity) }
+
+  before do
+    cs_accelerator
```
markjs, line 19: If the only purpose of this is to initialise the `cs_accelerator` value from the `let` then you could instead use `let!` which initialises it immediately then remove this whole `before` block.

spec/jobs/kick_off_emails_job_spec.rb
```diff
@@ -73,5 +75,22 @@
         end.not_to have_enqueued_job(ScheduleIBelongStudentSurveyPromptJob)
       end
     end
+
+    describe "with auto enrolment" do
+      it "should sent auto enrolled email" do
```
markjs, line 80: Think this is just a `sent` -> `send` typo 😄

spec/jobs/send_inactivity_emails_job_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+  let(:programme_activity_grouping_engagement) { create(:programme_activity_grouping, programme:, required_for_completion: 1, sort_key: 4) }
+  let(:programme_activity_engagement) { create(:programme_activity, programme:, activity: activity_engagement, programme_activity_grouping: programme_activity_grouping_engagement) }
+
+  before do
+    programme_activity_understanding
+    programme_activity_resources1
+    programme_activity_resources2
+    programme_activity_resources3
+    programme_activity_engagement
+  end
```
markjs, line 25: As per a comment above, this whole block could be removed if these are changed to use `let!` instead

spec/jobs/send_inactivity_emails_job_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+  describe "#perform" do
+    describe "users with no activities" do
+      let(:user_no_activities) { create(:user) }
+      let(:user_no_activities_upe) { create(:user_programme_enrolment, user: user_no_activities, programme:, created_at: 2.months.ago) }
+
+      before do
+        user_no_activities_upe
```
markjs, line 33: This is a before block for only one test case. Either using `let!` or moving it into the test case would be more consistent with other similar examples.

spec/lib/tasks/schedule_eligible_pending_enrolments_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+RSpec.describe "rake schedule_eligible_pending_enrolments", type: :task do
+  let(:programme) { create(:secondary_certificate) }
+  let(:enrolment) { create(:user_programme_enrolment, programme:) }
+
+  before do
+    enrolment
```
markjs, line 8: Another win for `let!` gang

markjs (approved):
A few small changes / efficiency gains but nothing big so will approve in anticipation of those changes being considered. Nice one 😄

PR #2049, opened 2024-05-24
Added course seed CP267

A-Wheeto:
Remove incorrect seed CY312

## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2710

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Added course seed CP267
- Removed incorrect course seed CY312

### DIFF
--- a/db/seeds/activities/stem_learning.rb
+++ b/db/seeds/activities/stem_learning.rb
@@ -2581,16 +2581,6 @@ Activity.find_or_initialize_by(stem_course_template_no: "477ce2ea-c20e-ef11-9f8a
   activity.programmes = [secondary_certificate]
 end.save!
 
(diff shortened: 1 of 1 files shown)


db/seeds/activities/stem_learning.rb
```diff
@@ -2602,3 +2602,15 @@
 
   activity.programmes = [secondary_certificate]
 end.save!
+
+Activity.find_by(stem_course_template_no: "690b99ca-1106-ef11-9f8a-002248c635a9")&.destroy
```
msquance-stem, line 2606: You should also remove the block that this is deleting as well, as otherwise we are seeding and then immediately destroying every time. We have the & to catch when the find doesn't match in future. 

I would also just add a comment as to why and what we are deleting, just for future reference

msquance-stem (changes requested):

msquance-stem (approved):
LGTM

PR #2052, opened 2024-05-29
Allow unenrolled users to re-enroll

croab:
## Status

* Review App link(s): Heroku doesn't seem happy to create one as of current so will populate when ready
* Closes [#2692](https://github.com/NCCE/teachcomputing.org-issues/issues/2692)

## Review progress:

- [x] Browser tested
- [ ] Tech review completed

## What's changed?

- Ensure the query for user enrolment checks that the enrolment is in a state of `:unenrolled` as well as whether the enrolment record is present in the first place.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -72,7 +72,10 @@ class User < ApplicationRecord
   end
 
   def enrolled_on_programme_pathway?(programme:, pathway:)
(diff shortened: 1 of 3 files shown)


GDTimi (commented):
The changes look good for catching the unenrolled enrollment scenario 👍

PR #2053, opened 2024-05-29
Adding extra testing for badges

msquance-stem:
Adding in testing specifically to check the difference between face 2 face activities versus community, after reports (from months back, so likely fixed previously) that users were getting badges from just doing community activities

## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2717

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Added specific testing for f2f versus community to ensure badges are allocated correctly based on activity type

### DIFF
--- a/spec/jobs/issue_badge_job_spec.rb
+++ b/spec/jobs/issue_badge_job_spec.rb
@@ -5,8 +5,12 @@ RSpec.describe IssueBadgeJob, type: :job do
   let!(:programme) { create(:cs_accelerator) }
   let(:badge) { create(:badge, :active, programme_id: programme.id, credly_badge_template_id: "00cd7d3b-baca-442b-bce5-f20666ed591b") }
   let(:user_programme_enrolment) { create(:user_programme_enrolment, user:, programme:) }
-  let!(:achievement) { create(:completed_achievement, user:) }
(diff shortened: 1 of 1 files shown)


spec/jobs/issue_badge_job_spec.rb
```diff
@@ -15,22 +19,30 @@
...
+      it "user is not enrolled" do
+        described_class.perform_now(achievement: f2f_achievement)
+        expect(Credly::Badge).not_to have_received(:issue)
+      end
+
+      it "when achievement is not F2F" do
+        user_programme_enrolment
+        described_class.perform_now(achievement: community_achievement)
+        expect(Credly::Badge).not_to have_received(:issue)
+      end
```
GDTimi, line 43: Just to clarify for my own understanding, are badges never awarded for community achievements - is there always a face to face component?

GDTimi (approved):
Badge test case updates look good for the comparative f2f / community paths.

spec/jobs/issue_badge_job_spec.rb
msquance-stem: Yeah the badges only should be issued for completing a CPD course (so face-to-face, remote, or online), but not for community activities.

PR #2056, opened 2024-05-30
Move "Implement selected key stage 3 Teach Computing Curriculum resources" from section 2 to 3

croab:
## Status

* Current Status: WIP
* Review App link(s): ...
* Closes [#2712](https://github.com/NCCE/teachcomputing.org-issues/issues/2712)

## Review progress:

- [x] Browser tested (although could do with further testing)
- [x] Tech review completed

## What's changed?

Moved the affected activity with title "Implement selected key stage 3 Teach Computing Curriculum resources" from its original programme activity grouping (which places it under section 2 in the UI) to its second programme activity grouping (placing it under 3 in the UI). Script then updates the `required_for_completion` attribute on the original grouping from 3 to 2.

- Run the method on `ProgrammeActivityGrouping.move_i_belong_section_from_2_to_3` to perform the above changes
- Remove this method in a separate PR

### DIFF
--- a/app/models/programme_activity_grouping.rb
+++ b/app/models/programme_activity_grouping.rb
@@ -58,6 +58,34 @@ class ProgrammeActivityGrouping < ApplicationRecord
     end.to_h
   end
 
+  # Move "Implement selected key stage 3 Teach Computing Curriculum resources" from section 2
+  # to section 3 and update the #required_for_completion value on the old grouping
+  def self.move_i_belong_section_from_2_to_3
+    # Perform the move of the activity to the new grouping
+    programme = Programme.find_by(slug: "i-belong")
+    raise "Programme could not be found" if programme.blank?
+
+    potential_groupings = programme.programme_activity_groupings.community
+    raise "There are more than 2 available groupings for the moved activity" unless potential_groupings.size == 2
+
+    activity = programme.activities.find_by(title: "Implement selected key stage 3 Teach Computing Curriculum resources")
+    raise "Activity could not be found" if activity.blank?
+
+    programme_activity = ProgrammeActivity.find_by(activity_id: activity.id)
+    raise "Programme activity could not be found" if programme_activity.blank?
+
+    original_grouping = programme_activity.programme_activity_grouping
+    raise "Current grouping could not be found" if original_grouping.blank?
+
+    new_grouping = potential_groupings.where.not(id: original_grouping.id).first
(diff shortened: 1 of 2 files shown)


croab:
@markjs - unsure whether it would be better to run the data changes in the console directly and then just merge the seed changes, or whether the whole thing can just be merged and then the migration method removed after?

app/models/programme_activity_grouping.rb
```diff
@@ -58,6 +58,34 @@ def users_completed(users:)
...
+    raise "Current grouping could not be found" if original_grouping.blank?
+
+    new_grouping = potential_groupings.where.not(id: original_grouping.id).first
+    raise "New grouping could not be determined" if new_grouping.blank?
+
+    programme_activity.update!(programme_activity_grouping_id: new_grouping.id)
+
+    # Update the original grouping to only require 2 completed activities
+    original_grouping.update!(required_for_completion: 2)
+  end
```
GDTimi, line 87: Running the method twice flipped it back, but both groups then had required_for_completion set to 2. Might need something to catch this, else can run into an outcome of updating the last group to "Increase girls' engagement by completing at least two activities".

GDTimi (changes requested):
Just left a comment on potentially updating the required_for_completion on the final activity group unintentionally. :)

app/models/programme_activity_grouping.rb
croab: Nice, instead of a guard I just ensured that it would run the same every time in the end!

GDTimi: Just re-ran the method, I don't think it has had quite the desired affect, as 'Implement selected key stage 3 Teach Computing Curriculum resources' isn't visible in either group for me afterwards (although the "at least one activity" has remained!).

GDTimi (changes requested):
The 'Implement selected key stage 3 Teach Computing Curriculum resources' activity wasn't visible for me after running the method, unless there is something else I'm doing on my end!

app/models/programme_activity_grouping.rb
croab: Should work OK now

GDTimi: Yep, that looks to have addressed it - nice one. 👍

GDTimi (approved):

PR #2058, opened 2024-05-30
Adding Career Videos to Unit and Lesson pages

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2673
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2707

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Created a new component to display videos on unit and lesson pages  - can display in column or row layout
- Updated graphql queries to get videos from curriculum
- Added video component to unit and lesson page
- Added tests for component and rendering of component on the views
- Small fix to padding on the aside on a lesson page to show correctly on mobile


app/components/curriculum_video_component/curriculum_video_component.html.erb
```diff
@@ -0,0 +1,24 @@
+<div class="video-component-<%= layout_class %>">
+  <div class="video-component-<%= layout_class %>__title">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-2">Career video to support the unit</h2>
+  </div>
+  <div class="video-component-<%= layout_class %>-content">
+    <div class="video-component-<%= layout_class %>-content__media">
+      <%= render YouTubeEmbedComponent.new(video_url: @video_url) %>
+    </div>
+    <div class="video-component-<%= layout_class %>-content--details">
```
msquance-stem, line 9: I wouldn't use the double dash between content and details, as -- implies details is a modifier, where I think you mean its a element, so should be double underscore?

Or as it contains other elements, it might just be best keeping it a - but only a single one, so its not confused as a modifier

app/components/curriculum_video_component/curriculum_video_component.scss
```diff
@@ -0,0 +1,82 @@
...
+
+    @include govuk-media-query($until: tablet) {
+      flex-direction: column;
+    }
+
+    &__media {
+      flex: 2;
+    }
+
+    &--details {
```
msquance-stem, line 29: As per comment on the other file, will need updating to reflect any change made in other file

msquance-stem (commented):
Few minor points about the CSS naming convention, but other than that it looks good

app/components/curriculum_video_component/curriculum_video_component.html.erb
A-Wheeto: Agreed. I'll change this to a single dash.

app/components/curriculum_video_component/curriculum_video_component.scss
A-Wheeto: I'll update this to reflect a single dash

msquance-stem (approved):
LGTM

PR #2060, opened 2024-06-03
Remove supplementary courses from CS accelerator pages

croab:
## Status

* Current Status: Ready for Review
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2716

## Review progress:

- [x] Browser tested
- [ ] Tech review completed

## What's changed?

- Removed the 'View additional courses?' section and much of the supplementary activities logic as it is no longer used anywhere else.

## Notes

- This PR assumes that the supplementary courses feature (really a pathway activity) is no longer needed as it is only used in the page section which has been designated for removal. I have, however, left the ability to denote a course as supplementary intact as this is used still by the `#recommended_activities` method of `Pathway`. Ideally, we'd just identify all `#recommended_activities` as just `pathway_activities` without the boolean `supplementary` `where` filter since supplementary courses are no longer used. Pending response to whether we still need to identify supplementary courses at all.


GDTimi (approved):
Changes look fine on the technical  & frontend sides, pending confirmation on the supplementary attribute removal.

msquance-stem:
We had been waiting on confirmation from programmes team that this was no longer needed, and we finally got it during the changes needed for April 1st 2025 contract changes

PR #2067, opened 2024-06-10
Fix to ensure title label can handle nil in admin

croab:
## What's changed?

Fix to ensure that title label can handle nil in admin interface for user programme enrolments


msquance-stem (approved):
LGTM

PR #2069, opened 2024-06-10
Fix for missing user on staging - guard added

croab:
Fix to ensure that staging user programme enrolments that are missing users do not throw exceptions


GDTimi (approved):
Guards in the methods look to cover it 👍

PR #2072, opened 2024-06-13
Text and image alignment on curriculum video component

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2724

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Moved the text and image on the details section into separate divs to maintain alignment and handle longer strings of text


msquance-stem (commented):
I am happy with the HTML, but looking at the CSS there is a lot of repetition between the two blocks, you could potentially simplify down but removing the layout class from the inner elements (e.g details), so the classes are the same regardless of the parent layout. As the layout only modifies the parent, its not for the content. That way then you can reduce the amount of css.

msquance-stem (approved):
Much better! LGTM

PR #2078, opened 2024-07-01
Remove reference to ActiveModel boolean cast in initializer

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2778

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Removed the use of the Rails Boolean casting method, instead we will just do a simple string comparison for the word true (lowercase) otherwise we will assume no bypass exists
- This is due to the call to this (outside of class in the omniauth initialiser) seems to intermittently cause the app to crash out, my guess is that sometimes it takes longer to boot rails and the initialiser is being run before the library code is loaded
- I have modified the second reference (inside the class) to keep consistency in the check methods, but that one doesnt appear to cause us an issue

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -32,7 +32,7 @@ module OmniAuth::Strategies
     end
 
     def callback_url
-      return super if ActiveRecord::Type::Boolean.new.cast(ENV["BYPASS_OAUTH"])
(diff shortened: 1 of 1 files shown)


A-Wheeto (approved):
LGTM

PR #2092, opened 2024-07-18
2783 sticky course filters look wonky on mobile

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2783

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Removed sticky behaviour on the course filter
- Added divs and additional CSS to move the "Filter courses" text outside and inside the box when viewing on desktop or mobile
- Changed "Filter courses" text to "More filters" inline with the mockup from Sam
- A few minor style tweaks

### DIFF
--- a/app/views/courses/_aside-filters.html.erb
+++ b/app/views/courses/_aside-filters.html.erb
@@ -1,9 +1,9 @@
 <% unless @course_filter.nil? %>
-<div class=" ncce-courses__filters-wrapper__title-desktop">
+<div class="ncce-courses__filters-wrapper__title-desktop">
   <h2 class="govuk-heading-s">Filter courses</h2>
 </div>
 <div class="ncce-aside ncce-aside--filters govuk-!-margin-top-2">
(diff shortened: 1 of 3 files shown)


app/views/courses/_aside-filters.html.erb
```diff
@@ -1,22 +1,27 @@
 <% unless @course_filter.nil? %>
-<h2 class="govuk-heading-s">Filter courses</h2>
+<div class="ncce-courses__filters-wrapper__title-desktop">
```
msquance-stem, line 2: Its not usual in the BEM model to have two lots of __, I would suggest it could be 
`ncce-courses__filter-wrapper-title--desktop`
and
`ncce-courses__filter-wrapper-title--mobile`
as the last bit is more of a modifier

msquance-stem (approved):
Up to you if you want to update the css class naming, but other than that it all looks good.

PR #2118, opened 2024-08-06
Course Credit Calculator

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2737

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Added new method in activity model to calculate credits for a course seed based on the duration in hours
- Logic to calculate credits if an existing activities programme is changed

### DIFF
--- a/spec/models/activity_spec.rb
+++ b/spec/models/activity_spec.rb
@@ -234,17 +234,17 @@ RSpec.describe Activity, type: :model do
 
   describe "#calculate_credits" do
     it "should use initial credit value" do
-      activity = create(:activity, credit: 50, duration_in_hours: 2.5)
+      activity = create(:activity, :with_duration, credit: 50)
       expect(activity.credit).to eq(50)
     end
 
     it "calculates the correct credits from duration in hours" do
-      activity = create(:activity, :activity_no_credits, duration_in_hours: 2.5)
+      activity = create(:activity, :with_duration, :activity_no_credits)
       expect(activity.credit).to eq(25)
     end
 
     it "should not change credit amount if no duration in hours" do
-      activity = create(:activity, credit: 50, duration_in_hours: nil)
+      activity = create(:activity, credit: 50)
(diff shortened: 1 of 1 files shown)


spec/factories/activities.rb
```diff
@@ -11,6 +11,7 @@
     always_on { false }
     retired { false }
     self_verification_info { nil }
+    duration_in_hours { 2.5 }
```
msquance-stem, line 14: I wonder if we should have this default to nil, as that currently the default behaviour we have on seeds, then maybe have a trait :with_duration?

spec/models/activity_spec.rb
```diff
@@ -231,4 +231,65 @@
       expect(activity.active_course?).to eq(false)
     end
   end
+
```
msquance-stem, line 234: We currently do not test that the credits are not changed if there is no duration in hours. As we use that as the primary criteria in the callback in the programme activity, would be good to test that behaviour

msquance-stem (commented):
Few little bits that would make this a little bit safer, but looking really good now

spec/factories/activities.rb
A-Wheeto: Just having a think about this. I need to test cs_accelerator cases with different durations (less than a day, more than one day).
Would it be cleaner to define the duration in hours when creating the activity for the individual test, rather than a trait? Otherwise I'd have to create multiple traits. 
Eg. `activity = create(:activity, :with_duration, credit: nil, programmes: [cs_accelerator_programme], duration_in_hours: 10)`

A-Wheeto: ...and have the default to nil as suggested.

spec/models/activity_spec.rb
A-Wheeto: I've added a test for this on line 246

PR #2119, opened 2024-08-07
Classmarker logic for timed out assessments

A-Wheeto:
## Status

* Ready for (front-end / tech) review
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2784

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Refactored class marker logic to account for timed out assessment attempts
- Assessment attempts can now enter a state of timeout
- Updates to method names to be more readable


app/jobs/update_user_assessment_attempt_from_class_marker_job.rb
```diff
@@ -6,6 +6,10 @@ def perform(test_id, user_id, percentage)
     assessment = Assessment.find_by!(class_marker_test_id: test_id)
     latest_attempt = assessment.latest_attempt_for(user:)
 
+    if percentage.nil?
```
msquance-stem, line 9: I don't think this is strictly needed, as I believe the only place this is called is from the Classmarker webhook, and that would never return without a percentage

app/models/user_programme_assessment.rb
```diff
@@ -19,7 +19,7 @@ def enough_activities_for_test?
   end
 
   def num_attempts
-    return 0 if @attempts.nil?
+    return 0 if @attempts.nil? || @attempts&.last&.current_state == StateMachines::AssessmentAttemptStateMachine::STATE_TIMEDOUT.to_s
```
msquance-stem, line 22: I think this could be written differently, as its a lot of `&.`, you might be able to count attempts that are in failed and complete, as the count is still going to include timed out tests.

@attempts could be changed to include only failed and complete, which then fixes the logic of can_take_test_at as well, as it just doesnt include the timed out ones

msquance-stem (commented):
There are a few things that could improve it, I have stuck comments against the code

spec/jobs/update_user_assessment_attempt_from_class_marker_job_spec.rb
```diff
@@ -12,6 +12,7 @@
   let(:assessment_attempt) { create(:assessment_attempt, user_id: user.id, assessment_id: assessment.id) }
   let(:programme_complete_counter) { create(:programme_complete_counter, programme_id: programme.id) }
   let(:passing_json_body) { File.read("spec/support/class_marker/passing_webhook.json") }
+  let(:timed_out_json_body) { File.read("spec/support/class_marker/timed_out_webhook.json") }
```
msquance-stem, line 15: I assume this file no longer exists? If so we can remove this.

msquance-stem (approved):
One very minor change needed, I will approve, but can you remove the un-needed line before merging

Good work!

spec/jobs/update_user_assessment_attempt_from_class_marker_job_spec.rb
A-Wheeto: Good spot Mike, I've removed this :)

PR #2136, opened 2024-08-21
PIB - Evidence modal character validation and user warning

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2809

Fixes the bug in this ticket:
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2801

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Additional javascript in community activity component added to check textareas for evidence and warns the user on the last section if they are missing evidence in one of the sections
 - Alongside this, added links to take the user back to that missing section
 - Some javascript added into the progress component so the warning only appears on the last section
- Added minimum character requirement to the evidence boxes - set to 50. When a user starts entering their evidence a text prompt will appear to show the user how many more characters are required to be a valid submission
- Small tweaks to the CSS on the modal for better layout on mobile view

### DIFF
--- a/app/components/community_activity_component.rb
+++ b/app/components/community_activity_component.rb
@@ -39,6 +39,6 @@ class CommunityActivityComponent < ViewComponent::Base
   end
 
   def minimum_character_requirement
-    @activity.programmes.map(&:minimum_character_required_community_evidence).max
+    @activity.programmes.collect(&:minimum_character_required_community_evidence).max
   end
 end
--- a/app/components/community_activity_component.rb
+++ b/app/components/community_activity_component.rb
@@ -37,4 +37,8 @@ class CommunityActivityComponent < ViewComponent::Base
     return default_option["slug"] == slug if default_option
     false
   end
+
+  def minimum_character_requirement
+    @activity.programmes.map(&:minimum_character_required_community_evidence).max
+  end
 end
--- a/app/components/community_activity_component/community_activity_component.html.erb
+++ b/app/components/community_activity_component/community_activity_component.html.erb
@@ -124,19 +124,24 @@
                             </ul>
                           <% end %>
                         </p>
-                        <textarea class="govuk-textarea govuk-!-margin-bottom-1" placeholder="Please provide us with evidence of the activity you delivered." data-community-activity-component-target="textarea"><%= @achievement&.evidence&.[](index) %></textarea>
-                        <div class="community-activity-component__completion-count">
+                        <textarea class="govuk-textarea govuk-!-margin-bottom-2" placeholder="Please provide us with evidence of the activity you delivered." data-community-activity-component-target="textarea"><%= @achievement&.evidence&.[](index) %></textarea>
+                        <div class="community-activity-component__completion-count govuk-!-margin-bottom-1">
                           <span class="govuk-body-s" data-community-activity-component-target="characterCountMessage"></span>
                         </div>
                       <% end %>
                     <% end %>
+                    <%= progress.with_warning_step do %>
+                      <div>
+                        <span class="govuk-body-m" data-community-activity-component-target="completionWarningMessage"></span>
+                      </div>
+                    <% end %>
 
                     <div class="community-activity-component__footer-actions">
                       <%= progress.with_back do %>
-                        Previous section
+                        Previous
                       <% end %>
                       <%= progress.with_continue do %>
-                        Next section
+                        Next
                       <% end %>
                       <%= render ModalComponent.new(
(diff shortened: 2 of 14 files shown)


A-Wheeto:
Feedback from Ollie after QA:
- Save draft is blocked unless text is entered into all three
- Could you also make the text at the bottom red when it indicated to a user they need to enter into all three

app/components/community_activity_component.rb
```diff
@@ -26,4 +26,8 @@ def reopen_button_text
     return "Add more evidence" if achievement_rejected?
     @achievement&.evidence.present? ? "Continue editing" : "Submit evidence"
   end
+
+  def minimum_character_requirement
+    @activity.programmes.map(&:minimum_character_required_community_evidence).max
```
msquance-stem, line 31: Just as an FYI, there is method called collect that I used when extracting data from an array of objects. It makes no difference which you use, as collect is just an alias for map. I use it as I think its clearer I am just 'collecting' the values, not doing something to them.

app/components/community_activity_component/community_activity_component.html.erb
```diff
@@ -122,17 +123,28 @@
...
-                          <textarea class="govuk-textarea" placeholder="Please provide us with evidence of the activity you delivered." data-community-activity-component-target="textarea"><%= @achievement&.evidence&.[](index) %></textarea>
                         </p>
+                        <textarea class="govuk-textarea govuk-!-margin-bottom-2" placeholder="Please provide us with evidence of the activity you delivered." data-community-activity-component-target="textarea"><%= @achievement&.evidence&.[](index) %></textarea>
+                        <div class="community-activity-component__completion-count govuk-!-margin-bottom-1">
+                          <span class="govuk-body-s" data-community-activity-component-target="characterCountMessage"></span>
+                        </div>
+                      <% end %>
+                    <% end %>
+
+                    <% if minimum_character_requirement != 0 %>
```
msquance-stem, line 134: Is this if statement needed, as we don't show the warning when its not needed? Might be in the future we want to use the warning step for other information that do not relate to the min char requirement?

app/components/progress_component/progress_component_controller.js
```diff
@@ -1,14 +1,15 @@
 import ApplicationController from "../../webpacker/javascript/controllers/application_controller"
 
 export default class ProgressComponentController extends ApplicationController {
-  static targets = ["counter", "back", "continue", "submit", "step"]
+  static targets = ["counter", "back", "continue", "submit", "step", "warning"]
 
   connect() {
     if (this.stepTargets.length == 0) throw new Error("Must have at least one step")
     this.currentStep = 0
     this.updateCounter()
```
msquance-stem, line 9: Could we just call updateSection in the connect method, rather than having this long list of methods?

app/webpacker/javascript/controllers/community_activity_component_controller.js
```diff
@@ -32,13 +33,41 @@ export default class extends ApplicationController {
...
+      if(element.value.length > 0 && element.value.length < this.minimumEvidenceLengthValue) {
+        this.characterCountMessageTargets[index].textContent = `Provide at least ${remainingCharacters} more characters`
+      } else {
+        this.characterCountMessageTargets[index].textContent = ""
+      }
+
+    };
+
+    this.textareaTargets.forEach((input, index) => {
+      input.addEventListener("input", () => this.updateCharacterCount(input, index))
```
msquance-stem, line 52: We should add a disconnect for this, to ensure the listener is removed when we remove the controller to keep the browser session tidy.

app/webpacker/javascript/controllers/community_activity_component_controller.js
```diff
@@ -51,7 +80,17 @@ export default class extends ApplicationController {
           this.saveDraftButtonTarget.setAttribute('disabled', '')
```
msquance-stem, line 80: Given what we discussed this morning about Draft button being enabled when they have some data, will this not cause problems as it going to disable that button when they have a missing value?

msquance-stem (changes requested):
Few couple of minor points, and a question about the logic for the save as draft button, given what we discussed in standup this morning.

Rest looks good though

app/webpacker/javascript/controllers/community_activity_component_controller.js
msquance-stem: Ignore this, just me being stupid and not noticing the code fold in the github review

msquance-stem (approved):
LGTM, good work!

PR #2148, opened 2024-08-29
New Enrichment Pages

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2741

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- New enrichment page all managed via strapi
- Includes new text with asides component from strapi dynamic zones


app/components/cms_text_with_asides_component/cms_text_with_asides_component.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= render CmsRichTextBlockComponent.new(blocks: @blocks, with_wrapper: false) %>
+  <% end %>
+
+  <%= row.with_column("one-third") do %>
+
+    <% @aside_models.each do |aside| %>
+      <% if aside[:model] %>
+        <%= render aside[:model].render %>
+      <% else %>
+        <!-- missing aside <%= aside[:slug] %> -->
```
A-Wheeto, line 12: Does this comment need removing?

A-Wheeto (changes requested):
Looks great Mike, I noticed a comment that I assume needs to be removed?

app/components/cms_text_with_asides_component/cms_text_with_asides_component.html.erb
msquance-stem: Good question, I had it on the references to the hard coded asides we had in the dashboard, but I guess in the strapi version they should never be missing. I will remove it.

A-Wheeto (approved):
LGTM

PR #2151, opened 2024-09-16
Strapi horizontal card component

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2831

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Created new CMS horizontal card component which allows optional image, icon row, ribbon and image link
- New CMS model for component within a dynamic zone in Strapi
- New CMS model for icon block
- New component parameter factory to populate data from Strapi
- Created global css for colour theme from Strapi and a helper method to apply to components
- Appropriate tests and previews for the new components


app/components/cms_horizontal_card_component/cms_horizontal_card_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<%= render GovGridRowComponent.new(padding: {top: 6, bottom: 0}) do |row|  %>
+  <%= row.with_column("full") do %>
+    <%= content_tag :div, class: [
```
msquance-stem, line 3: I would recommend moving the class generation up to the component, makes the view cleaner to read. You could have a method called wrapper_classes, that returns an array of classes

app/components/cms_horizontal_card_component/cms_horizontal_card_component.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      @colour_theme ? cms_colour_theme(@colour_theme, "left") : nil].compact do %>
+
+      <div class="horizontal-card-component__content">
+        <h1 class="govuk-heading-m"><%= @title %></h1>
+        <%= render CmsRichTextBlockComponent.new(blocks: @body_blocks, with_wrapper: false) %>
+        <%= render @icon_block.render if @icon_block%>
+      </div>
+      <% if @image %>
+        <div class="horizontal-card-component__image">
+            <%= @image_link&.then { |link| link_to(link) { render CmsImageComponent.new(@image) } } || render(CmsImageComponent.new(@image)) %>
```
msquance-stem, line 14: Could we simplify this by having the link as part of the CmsImageComponent as an optional parameter?

As we have a few places where we might link an image, we could standardise it, rather than having to do it in lots of different components

app/components/cms_icon_block_component/cms_icon_block_component.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="icon-wrapper">
+  <% @icons.each do |icon| %>
+    <div class="icon-wrapper__block">
+      <div class="icon-wrapper__image">
+        <%= render CmsImageComponent.new(icon.image) %>
+      </div>
+      <div class="icon-wrapper__text">
+        <p class="govuk-body-s govuk-!-margin-bottom-0"><%= icon.text %></p>
```
msquance-stem, line 8: Could we use a span here, then we dont need to mess with the margin botton?

app/helpers/cms_helper.rb
```diff
@@ -14,4 +15,8 @@ def cms_image_url(image)
   def cms_image(image, class: nil)
     image_tag(cms_image_url(image), alt: image.alt, class:)
   end
+
+  def cms_colour_theme(colour, side)
```
msquance-stem, line 19: Should we add class to the name of this method, just to make it clear its returning a class name?

app/services/cms/providers/strapi/mocks/image.rb
```diff
@@ -21,7 +21,7 @@ def self.generate_data
             {
               name: filename,
               alternativeText: Faker::Lorem.sentence,
-              caption: Faker::Lorem.sentence,
+              caption: "",
```
msquance-stem, line 24: Can we add caption as a keyword param on generate_data, and make it nil default rather than an empty string

previews/components/cms_icon_block_component_preview.rb
```diff
@@ -0,0 +1,5 @@
+class CmsIconBlockComponentPreview < ViewComponent::Preview
+  def default
+    # TODO ADD COMPONENT PREVIEW
```
msquance-stem, line 3: Need to add the component preview here

msquance-stem (commented):
Looks good for the most part, well done, I have just suggested a few areas that you might want to review. Have a look through them and see what you think, happy to discuss.

msquance-stem (approved):
LGTM, thanks for making those changes

PR #2166, opened 2024-09-23
Add resource cards to Strapi

A-Wheeto:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?


app/components/cms_resource_card_component/cms_resource_card_component.scss
```diff
@@ -0,0 +1,41 @@
+.cms-resource-card {
+  &_wrapper {
```
msquance-stem, line 2: It appears cms-resource-card as a class is never used, and _wrapper is just appended on the end. Do we need the wrapper, can we not just have the `cms-resource-card` be the main class?

app/services/cms/providers/strapi/factories/component_factory.rb
```diff
@@ -27,6 +27,26 @@ def self.process_component(strapi_data)
...
+            end
+          end
+
+          def self.resource_card_block(strapi_data)
+            strapi_data.map do |card_data|
+              Models::ResourceCard.new(
+                title: card_data[:title],
+                icon: card_data.dig(:icon, :data) ? ModelFactory.to_image(card_data[:icon][:data][:attributes]) : nil,
+                colour_theme: card_data.dig(:colourTheme, :data) ? card_data[:colourTheme][:data][:attributes][:name] : nil,
+                body_text: card_data[:bodyText],
```
msquance-stem, line 46: It might be worth wrapping these bits that use the Rich Block from strapi in `ModelFactory.to_content_block` as this auto processes any images they insert into the content block. It might not be a case that they are going to put images in, but without it, it will crash.

msquance-stem (approved):
LGTM

PR #2175, opened 2024-10-10
Implement new certificates overview dashboard

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2845

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Refactored existing dashboard to the new design and remove redundant partials, css and tests
- Created a dashboard certificate view component
- Create a dashboard course component


app/controllers/dashboard_controller.rb
```diff
@@ -4,12 +4,36 @@ class DashboardController < ApplicationController
   after_action :discourage_caching
 
   def show
-    @incomplete_achievements = current_user.achievements.not_in_state(:dropped, :complete).with_courses.order("created_at DESC")
-    @completed_achievements = current_user.achievements.in_state(:complete).with_courses.order("updated_at DESC")
+    @incomplete_achievements = incomplete_achievements
+    @completed_achievements = completed_achievements
+    @enrolled_certificates, @unenrolled_certificates = enrolments
+  end
+
```
msquance-stem, line 11: I would make all these following methods private by adding a `private` at this point, otherwise they could be called directly in the future (if someone adds a route), but these are routes, just internal data handlers

app/models/programme.rb
```diff
@@ -191,4 +191,12 @@ def notification_link
   def minimum_character_required_community_evidence
     0
   end
+
+  def school_achievement?
```
msquance-stem, line 195: Would it be better to have an `achievement_type` method, that has a raises NotImplementedError in programme, then override it in each of the specific classes to return either `:school` or `:individual`? 

I think that would give us more flexibility in the future

app/views/dashboard/show.html.erb
```diff
@@ -1,38 +1,71 @@
...
-          <%= render 'dashboard/certificates/i_belong', programme: Programme.i_belong %>
 
-				</div>
-			<%= render 'dashboard/courses' %>
-		</div>
-	</div>
-</div>
+<%= render CmsPageTitleComponent.new(title: "Progress overview", intro_text: nil) %>
+
+<% if @enrolled_certificates&.any? %>
```
msquance-stem, line 5: Is the & needed, as we define this always in the controller? As is it not an empty array if there are none, or does it come back as nil?

app/views/dashboard/show.html.erb
```diff
@@ -1,38 +1,71 @@
...
+  <%= render GovGridRowComponent.new(background_color: "light-grey", padding: {top: 7, bottom: 1}) do |row| %>
+    <%= row.with_column("two-thirds") do %>
+      <h1 class="govuk-heading-l govuk-!-margin-bottom-4">Enrolled certificates</h1>
+      <% @enrolled_certificates.each do |certificate| %>
```
msquance-stem, line 9: Same as other comment

app/views/dashboard/show.html.erb
```diff
@@ -1,38 +1,71 @@
...
+        Choose your certificate
+      <% end %>
+    </h1>
+    <% @unenrolled_certificates.each do |certificate| %>
+      <%= render DashboardCertificateComponent.new(certificate:, button_text: "Find out more", button_color: "ncce-button--white") %>
+    <% end %>
+  <% end %>
+<% end %>
+
+<% if @incomplete_achievements&.any? %>
```
msquance-stem, line 37: & again, is it needed?

msquance-stem (commented):
Looks good, just a few things that need addressing

app/components/dashboard_course_component/dashboard_course_component.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="dashboard-course-component">
+  <%= link_to @achievement.activity.title, course_path(id: @achievement.activity&.stem_activity_code, name: @achievement.activity.slug),
```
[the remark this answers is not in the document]
A-Wheeto: As discussed we don't need the & here and I will remove it.

app/components/dashboard_course_component/dashboard_course_component.scss
```diff
@@ -5,43 +5,41 @@
...
-      margin-bottom: 0;
-    }
+  .dashboard-course-component__format {
+    display: flex;
+    padding-top: 5px;
 
-    &:last-child {
-      margin-bottom: 0;
+    span {
+      font-size: 16px;
```
msquance-stem, line 33: You could use the govuk mixin here to set this, as it brings in other settings related to typography.

`@include govuk-font($size: 16);`

msquance-stem (commented):
Looking good, just one little point, I will wait till you finished and give it a final review and approval, but good work

app/components/dashboard_course_component/dashboard_course_component.scss
A-Wheeto: Good spot! Will update this.

spec/factories/achiever/course_delegate.rb
```diff
@@ -0,0 +1,24 @@
+FactoryBot.define do
+  factory :course_delegate, class: "Achiever::Course::Delegate" do
+    course_occurence_no { Faker::Number.number(digits: 8) }
+    course_template_no { Faker::Number.number(digits: 8) }
+    is_fully_attended { "false" }
+    online_cpd { false }
+    progress { "157420003" }
```
msquance-stem, line 7: Not essential, but we could add some traits for the different progress states, just to make it easier to test in the future?

msquance-stem (approved):
Looking good now, I have just added one comment, not essential, just a nice to have to save us some work in the future potentially.

Good work

msquance-stem (approved):
LGTM

PR #2191, opened 2024-10-22
Crup Setup

msquance-stem:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?


app/models/user.rb
```diff
@@ -6,16 +6,16 @@ class User < ApplicationRecord
   validates :last_name, presence: true
   validates :stem_achiever_contact_no, presence: true
   validates :stem_credentials_access_token, presence: true
-  validates :stem_credentials_refresh_token, presence: true
+  # validates :stem_credentials_refresh_token, presence: true
```
A-Wheeto, line 9: Should we still be validating the refresh token here?

msquance-stem: We no longer get a refresh token. It used to come with the login, we stored it, and never used it.

I had commented it out during testing, will just delete

A-Wheeto (approved):
Apart from those 2 bits of code to remove, LGTM!

PR #2197, opened 2024-10-25
Implement primary certificate enrolment confirmation

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2861

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Created enrolment confirmation component
- Added method onto programme model that defaults confirmation to false unless specified on the programme itself as true
- Added relevant tests and previews

### DIFF
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -2,8 +2,9 @@
 
 class EnrolmentConfirmationComponent < ViewComponent::Base
   delegate :auth_url, to: :helpers
+  attr_reader :button_text
 
-  def initialize(programme:, current_user:, full_width: true, button_text: nil)
+  def initialize(programme:, current_user:, full_width: true, button_text: "Enrol")
     @current_user = current_user
     @programme = programme
     @full_width = full_width
@@ -14,9 +15,5 @@ class EnrolmentConfirmationComponent < ViewComponent::Base
     classes = ["govuk-button button"]
     classes << "button--full-width" if @full_width
     classes
-  end
-
-  def button_text
-    @button_text || "Enrol"
   end
 end
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -3,9 +3,16 @@
 class EnrolmentConfirmationComponent < ViewComponent::Base
   delegate :auth_url, to: :helpers
 
-  def initialize(programme:, current_user:, enrol_text: nil)
+  def initialize(programme:, current_user:, full_width: true, enrol_button_text: nil)
     @current_user = current_user
     @programme = programme
-    @enrol_text = enrol_text
+    @full_width = full_width
+    @enrol_button_text = enrol_button_text
+  end
+
+  def button_classes
+    classes = ["govuk-button button"]
+    classes << "button--full-width" if @full_width
+    classes
   end
 end
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -15,4 +15,8 @@ class EnrolmentConfirmationComponent < ViewComponent::Base
     classes << "button--full-width" if @full_width
     classes
   end
+
+  def button_text
+    @button_text || "Enrol"
+  end
 end
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class EnrolmentConfirmationComponent < ViewComponent::Base
-  delegate :auth_url, to: :helpers
   attr_reader :button_text
+
+  delegate :auth_url, to: :helpers
 
   def initialize(programme:, current_user:, full_width: true, button_text: "Enrol")
     @current_user = current_user
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -5,11 +5,12 @@ class EnrolmentConfirmationComponent < ViewComponent::Base
 
   delegate :auth_url, to: :helpers
 
-  def initialize(programme:, current_user:, full_width: true, button_text: "Enrol")
+  def initialize(programme:, current_user:, full_width: true, button_text: "Enrol", pathway: nil)
     @current_user = current_user
     @programme = programme
     @full_width = full_width
     @button_text = button_text
+    @pathway = pathway
   end
 
   def button_classes
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -3,11 +3,11 @@
 class EnrolmentConfirmationComponent < ViewComponent::Base
   delegate :auth_url, to: :helpers
 
-  def initialize(programme:, current_user:, full_width: true, enrol_button_text: nil)
+  def initialize(programme:, current_user:, full_width: true, button_text: nil)
     @current_user = current_user
     @programme = programme
     @full_width = full_width
-    @enrol_button_text = enrol_button_text
+    @button_text = button_text
   end
 
   def button_classes
--- a/app/components/enrolment_confirmation_component.rb
+++ b/app/components/enrolment_confirmation_component.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class EnrolmentConfirmationComponent < ViewComponent::Base
+  delegate :auth_url, to: :helpers
+
+  def initialize(programme:, current_user:, enrol_text: nil)
+    @current_user = current_user
+    @programme = programme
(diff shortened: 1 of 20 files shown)


app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
```diff
@@ -0,0 +1,28 @@
...
+        title: "Enrol on the #{@programme.certificate_name}",
+        reopen_button_text: (@button_text || "Enrol"),
+        show_corner_decoration: false,
+        reopen_button_class: ("enrolment-confirmation-component__button--full-width" if @full_width)
+        ) do |component| %>
+        <% component.with_body do %>
+          <div class="enrolment-confirmation-component__wrapper">
+            <div>
+              <h2 class="govuk-heading-s">Ready to enrol?</h2>
+              <p class="govuk-body">Confirm your enrolment on the certificate, and then let"s find you the perfect training course to get started with.</p>
```
msquance-stem, line 14: You have a double quote in lets

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
```diff
@@ -0,0 +1,28 @@
+<% if @current_user %>
+  <% if @programme.enrolment_confirmation_required? %>
+    <div class="enrolment-confirmation-component">
+      <%= render ModalComponent.new(
+        title: "Enrol on the #{@programme.certificate_name}",
+        reopen_button_text: (@button_text || "Enrol"),
```
msquance-stem, line 6: Could we have a method in the component called button text? As we do this || twice, be nice to have it in one place.

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
```diff
@@ -0,0 +1,28 @@
...
+              <p class="govuk-body">Confirm your enrolment on the certificate, and then let"s find you the perfect training course to get started with.</p>
+            </div>
+            <div>
+              <%= link_to "Comfirm my enrolment", @programme.enrol_path(user_programme_enrolment: { user_id: @current_user.id, programme_id: @programme.id }), method: :post, class: "govuk-button button", role: "button", draggable: "false" %>
+            </div>
+          </div>
+        <% end %>
+      <% end %>
+    </div>
+  <% else %>
```
msquance-stem, line 23: We have nothing in here to deal with a logged in user who is already enrolled?

I think current versions swap to a "view my dashboard" button when they are already enrolled?

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
```diff
@@ -0,0 +1,28 @@
...
+    <%= link_to (@button_text || "Enrol"), @programme.enrol_path(user_programme_enrolment: { user_id: @current_user.id, programme_id: @programme.id }), method: :post, class: button_classes, role: "button", draggable: "false" %>
+  <% end %>
+<% else %>
+  <%= link_to "Log in", auth_url, method: :post, class: button_classes, role: "button", draggable: "false" %>
```
msquance-stem, line 27: Do we need the draggable? Not sure what its needed for?

msquance-stem (commented):
Looking pretty good, I have added a few comments

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
A-Wheeto: Good spot, I didn't consider that scenario. I'll double check what we currently do and add in some logic to handle that.

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
A-Wheeto: I copied this link_to straight from one of the existing pages, now you point it out I'm not sure why draggable is in there either. I'll remove it.

spec/rails_helper.rb
```diff
@@ -73,6 +74,13 @@
   config.include FeatureFlagHelper
   config.include ViewComponent::TestHelpers, type: :component
   config.include Capybara::RSpecMatchers, type: :component
+  config.include ViewComponent::SystemTestHelpers, type: :component_sys_test
+
+  config.before(:each, type: :component_sys_test) do
```
msquance-stem, line 79: Might be worth adding a comment linking the issue were we found this fix in here, just incase people come across it and wonder why.

app/components/enrolment_confirmation_component.rb
```diff
@@ -0,0 +1,22 @@
...
+    @button_text = button_text
+  end
+
+  def button_classes
+    classes = ["govuk-button button"]
+    classes << "button--full-width" if @full_width
+    classes
+  end
+
+  def button_text
```
msquance-stem, line 19: We could replace this method by just making button_text default to "Enrol" in the constructor

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
```diff
@@ -0,0 +1,32 @@
...
+            </div>
+            <div>
+              <%= link_to "Confirm my enrolment", @programme.enrol_path(user_programme_enrolment: { user_id: @current_user.id, programme_id: @programme.id }), method: :post, class: "govuk-button button", role: "button" %>
+            </div>
+          </div>
+        <% end %>
+      <% end %>
+    </div>
+  <% else %>
+    <% if @programme.user_enrolled?(@current_user) %>
```
msquance-stem, line 24: Is the logic here right?

What happens if I am logged in, and enrolled on a programme that requires enrolment confirmation? Should the check for if they enrolled be before checking programme type?

previews/components/enrolment_confirmation_component_preview.rb
```diff
@@ -0,0 +1,25 @@
+class EnrolmentConfirmationComponentPreview < ViewComponent::Preview
+  def default
+    render(EnrolmentConfirmationComponent.new(
+      current_user: User.last.presence,
```
msquance-stem, line 4: Do we need presence on these user calls?

We also have no preview for the user not being logged in?

spec/components/enrolment_confirmation_component_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    it "renders the button text" do
+      expect(page).to have_text("Click here to enrol")
+    end
+
+    it "does not apply the full width button classes" do
+      expect(page).to_not have_css(".enrolment-confirmation-component__button--full-width")
+    end
+  end
+
+  context "when the user is already enrolled" do
```
msquance-stem, line 66: In relation to my comment early, we have only tested when the user in enrolled on a non-confirmation programme, not a one that also requirements confirmation

spec/system/components/enrolment_confirmation_component_spec.rb
msquance-stem, line 1: Do we need to test the non-logged in version and non confirmation versions in system tests too, or are they covered enough in the component test?

msquance-stem (commented):
Might be not reading the logic correctly, but I dont think we have account for an enrolled user on a confirmation required programme?

app/components/enrolment_confirmation_component/enrolment_confirmation_component.html.erb
A-Wheeto: Yes, good spot. The enrolment state needs to be checked before the programme.

spec/system/components/enrolment_confirmation_component_spec.rb
A-Wheeto: Yes, would be good to test the modal doesn't open for these versions.

app/views/certificates/pathways/show.html.erb
```diff
@@ -94,7 +94,16 @@
...
       <div class="govuk-grid-column-one-third">
-        <%= render PathwayEnrolComponent.new(programme: @programme, pathway: @pathway, current_user: @current_user) %>
+        <%= render AsideComponent.new(title: 'How to enrol') do |component| %>
+          <% component.with_body do %>
+            <ol class="govuk-body-s">
+              <% @pathway.enrol_copy.each do |line| %>
+                <li><%= line.html_safe %></li>
+              <% end %>
+            </ol>
+            <%= render EnrolmentConfirmationComponent.new(current_user:, programme: Programme.primary_certificate) %>
```
msquance-stem, line 104: I don't think this should be hard coded to programme primary, as programme is variable in this file so it can be used by multiple programmes

msquance-stem (commented):
Can we just check the impact of changing the pathway file as its still being used by secondary, and deal with the hard coded programme

msquance-stem (approved):
LGTM

PR #2227, opened 2024-11-25
Converting blank achiever_organisation_id to nil

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2910

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

In the new CRUP system, the achiever_organisation_id returns as a blank string, rather than a null. This should be null to be consistent with older implementations. This will convert blanks to nil.

Need to process anyone who has a blank organisation ID to nil after deployment (to deal with anyone in the interim, and fix the data teams reports)


A-Wheeto (approved):
LGTM

PR #2230, opened 2024-11-26
Dashboard bug fixes

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2911

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Fixed margin on ncce white button
- Fixed logic on enrolment text dependant on enrolment status
- Fix to pin icon clipping

### DIFF
--- a/app/components/dashboard_course_component/dashboard_course_component.html.erb
+++ b/app/components/dashboard_course_component/dashboard_course_component.html.erb
@@ -29,9 +29,7 @@
       <% end %>
     </div>
     <div class="dashboard-course-component__format">
-      <div class="dashboard-course-component__format--icon">
-        <span class="govuk-body <%= course_icon_class(@achievement.activity) %>"><%= t("shared.course_types.#{course_category(@achievement.activity)}") %></span>
-      </div>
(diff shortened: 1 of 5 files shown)


app/components/dashboard_course_component/dashboard_course_component.html.erb
```diff
@@ -30,10 +30,7 @@
     </div>
     <div class="dashboard-course-component__format">
       <div class="dashboard-course-component__format--icon">
-        <span class="govuk-body-s <%= course_icon_class(@achievement.activity) %>"></span>
-      </div>
-      <div class="dashboard-course-component__format--text">
```
msquance-stem, line 35: Do we no longer need the format--text class? If so, do we want the modifier --icon on the class, and could it just be a format div?

msquance-stem (approved):
LGTM - There is a little bit of HTML that might be able to be simplified but I will leave that up to you

app/components/dashboard_course_component/dashboard_course_component.html.erb
A-Wheeto: Ah yes, I'll reduce this down to just a format div

msquance-stem (approved):
LGTM

PR #2235, opened 2024-12-02
Implement new teach computing primary dashboard

A-Wheeto:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?


app/components/community_evidence_submission_modal_component.rb
```diff
@@ -0,0 +1,43 @@
+class CommunityEvidenceSubmissionModalComponent < ViewComponent::Base
+  def initialize(activity:, achievement: nil, button_class: nil)
+    @activity = activity
+    @achievement = achievement
+    @button_class = button_class
+  end
+
+  def achievement_complete?
```
msquance-stem, line 8: I wonder if we could move this method into the achievement model, as we have it in two different components?

Call it `community_achievement_complete?` or something like that?

app/components/primary_dashboard_programme_community_activity_group_component.rb
```diff
@@ -0,0 +1,32 @@
...
+    super(aside_sections:)
+    @title = title
+    @programme_activity_group = programme_activity_group
+  end
+
+  def before_render
+    @current_user = current_user
+  end
+
+  def user_programme_activities
```
msquance-stem, line 22: Would it be worth just having these as separate variables, rather than passing as a hash. Would mean refactoring the component its going into, but might improve readability on the component?

app/components/primary_dashboard_programme_community_activity_group_component.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class PrimaryDashboardProgrammeCommunityActivityGroupComponent < CmsWithAsidesComponent
+  delegate :current_user, to: :helpers
+
+  def initialize(title:, programme_activity_group:, current_user:, aside_slug: nil)
```
msquance-stem, line 6: When we call this, we pass in the aside slug and title, which are just coming from the programme_acitivty_group. Could we default these to nil and if they are nil get the value from the model, rather than having to pass them in if we are using the values on the model.

Do we need the option to override them, if not we can just not bother with the nil parameters?

app/components/user_programme_course_bookings_with_asides_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    @programme = programme
+  end
+
+  def before_render
+    @current_user = current_user
+
+    in_progress_achievements = current_user.achievements.in_state(:in_progress, :enrolled).with_courses.order("created_at DESC")
+    complete_achievements = current_user.achievements.in_state(:complete).with_courses.order("created_at DESC")
+
+    @courses =
```
msquance-stem, line 23: Similar to my other comment, would this be more readable if these were separate variables, rather than being a hash?

app/components/user_programme_course_bookings_with_asides_component.rb
```diff
@@ -0,0 +1,46 @@
...
+    else
+      if activity.remote_delivered_cpd
+        "icon-remote"
+      else
+        "icon-map-pin"
+      end
+    end
+  end
+
+  def course_duration(course)
```
msquance-stem, line 43: Could this be renamed as course instance/template, as its not returning the duration, its returning the template?

app/components/user_programme_course_bookings_with_asides_component/user_programme_course_bookings_with_asides_component.html.erb
```diff
@@ -0,0 +1,79 @@
+<%= render GovGridRowComponent.new(padding: {top: 0, bottom: 7}) do |row| %>
+  <%= row.with_column("two-thirds") do %>
+    <div class="user-programme-course-bookings-with-asides">
+      <div class="user-programme-course-bookings-with-asides__header">
+        <% if @programme.programme_activity_groupings.not_community.first.user_complete?(@current_user) %>
```
msquance-stem, line 5: Should this just call user_complete? on the group rather than the activity?

app/components/user_programme_course_bookings_with_asides_component/user_programme_course_bookings_with_asides_component.html.erb
```diff
@@ -0,0 +1,79 @@
...
+          <div>
+            <%= image_pack_tag('media/images/icons/activity-complete-icon.svg') %>
+          </div>
+        <% end %>
+        <div>
+          <h2 class="govuk-heading-m govuk-!-margin-bottom-0">Step one: CPD courses</h2>
+        </div>
+      </div>
+
+      <% if @courses[:in_progress].present? %>
```
msquance-stem, line 15: These next two loops are the same I believe?

If so could we find some way to combine then to reduce duplication

app/controllers/achievements_controller.rb
```diff
@@ -6,7 +6,12 @@ def create
     @achievement = current_user.achievements.build(achievement_params)
 
     if @achievement.save && @achievement.transition_to(:drafted)
```
msquance-stem, line 8: Should we move the transition from this bit down into the if statement, that then removes my other comment about the need to go from drafted to enrolled?

app/components/primary_dashboard_community_activity_component.rb
```diff
@@ -0,0 +1,29 @@
...
+  private
+
+  def available_activities
+    @available_activities ||= @programme_activities.reject do |pa|
+      chosen_activity_ids.include?(pa.activity.id)
+    end
+  end
+
+  def chosen_activity_ids
+    (@user_programme_activities[:incomplete] + @user_programme_activities[:complete])
```
msquance-stem, line 26: As per my other comments, this could this be more readable if these are two different variables rather than a hash?

msquance-stem (changes requested):
Mostly looks good, few areas where you can improve readability by adding some extra variables.
Also one of two places where we can reduce some of the duplication as well.
I have not been through all the tests yet, as they might change based my suggestions

app/components/user_programme_course_bookings_with_asides_component/user_programme_course_bookings_with_asides_component.html.erb
A-Wheeto: Good spot, definitely not a dry approach here

app/components/primary_dashboard_programme_community_activity_group_component.rb
msquance-stem, line 1: In this file we only use the completed activities to check if they are complete. Would it not be easier to call user_complete? on the programme_activity group as currently you only check the required_count, but in future we might have different requirements, and the PAG exposes a method so we dont have to worry about what makes it complete.

Then you can move the completed/incomplete activities query down to the child componet, rather than passing them in? See my notes on that file for justification

app/components/primary_dashboard_community_activity_component/primary_dashboard_community_activity_component.html.erb
```diff
@@ -0,0 +1,69 @@
...
+        </div>
+      <% end %>
+    </div>
+  <% end %>
+
+  <% if @incomplete_activities.present? %>
+    <div class="primary-dashboard-community-activity__activity" data-primary-dashboard-community-activity-component-target="activityDetails">
+      <h2 class="govuk-heading-s">Complete your chosen <%= "activity".pluralize(@incomplete_activities.count) %></h2>
+      <% @incomplete_activities.each do |upa| %>
+        <% achievement = (@current_user.achievements.find { |a| a.activity_id == upa.activity_id }) %>
```
msquance-stem, line 33: In here we pull out the achievement for the activity. But as the original query is from the user context, we could already have that in memory, saving the need for two iterations.

If you move the methods for complete and incomplete down to here, we can optimise the queries and reduce the need to look up the achievements.

app/components/user_programme_course_bookings_with_asides_component.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def completed_courses
+    @current_user.achievements
+      .in_state(:complete)
+      .with_courses.order("created_at DESC")
+      .belonging_to_programme(@programme)
+  end
+
+  def course_icon_class(activity)
```
msquance-stem, line 42: We don't need to redefine this, as we have it in a helper method, in fact we seem have multiple versions of it 
https://github.com/NCCE/teachcomputing.org/blob/d811cae9eea9258b60829ea757d08eac01de6569/app/helpers/activities_helper.rb#L13
https://github.com/NCCE/teachcomputing.org/blob/d811cae9eea9258b60829ea757d08eac01de6569/app/helpers/dashboard_helper.rb#L10

app/controllers/achievements_controller.rb
```diff
@@ -5,8 +5,14 @@ class AchievementsController < ApplicationController
...
 
-    if @achievement.save && @achievement.transition_to(:drafted)
-      flash[:notice] = "'#{@achievement.activity.title}' progress has been saved"
+    if @achievement.save
+      if params[:enrol]
+        @achievement.transition_to(:enrolled)
+        flash[:notice] = "'#{@achievement.activity.title}' has been added to your activity list"
+      else
+        flash[:notice] = "'#{@achievement.activity.title}' progress has been saved"
+        @achievement.transition_to(:drafted)
```
msquance-stem, line 14: Can we just swap these two lines round, so the flash is set second like the block above?

msquance-stem (changes requested):
Looking much better, a few little things to change.

I think we can still improve the community sections though, and save the need to pass the parameters through and improve the lookup to save having to do multiple iterations

app/components/primary_dashboard_community_activity_component/primary_dashboard_community_activity_component.html.erb
```diff
@@ -0,0 +1,68 @@
...
+            <div class="primary-dashboard-community-activity__selection-rejected">
+              <i class="icon-feedback"></i>
+              <span class="primary-dashboard-community-activity__selection-rejected-text govuk-body-s">
+                Please provide more evidence to this activity for review.
+              </span>
+            </div>
+          <% end %>
+          <div class="primary-dashboard-community-activity__activity-details-title">
+            <div>
+              <p class="govuk-body-m"><%= upa.activity.title %></p>
```
msquance-stem, line 44: I think we need a sanitize on the title, as the primary updates have links in the title.

app/components/primary_dashboard_community_activity_component/primary_dashboard_community_activity_component.html.erb
```diff
@@ -0,0 +1,68 @@
...
+          <p class="govuk-body-s"><%= upa.activity.public_copy_description.html_safe %></p>
+          <%= render CommunityEvidenceSubmissionModalComponent.new(activity: upa.activity, achievement: upa) %>
+        </div>
+      <% end %>
+    </div>
+  <% end %>
+
+  <div class="primary-dashboard-community-activity__selection-wrapper">
+    <%= select_tag :activity, options_for_select(
+      available_activities.map { |pa| [pa.activity.title, pa.activity.id ]}
```
msquance-stem, line 63: Might need a strip_tags on the title in here for the same reason as above. Might be that we also strip_tags on the previous one, as that might be cleaner on the dashboard?

msquance-stem (approved):
LGTM. Well done!

PR #2265, opened 2025-01-15
CMS Factory refactor

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2934

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Refactored factories into new base_factory, button_factory, blocks_factory, content_block_factory for better readability
- Refactored some cms component tests
- General tidy up of the factories


app/services/cms/models/text_block.rb
```diff
@@ -3,7 +3,7 @@ module Models
     class TextBlock
       attr_accessor :blocks, :with_wrapper
 
-      def initialize(blocks:, with_wrapper: false, **options)
```
msquance-stem, line 6: We mostly pass false into here, so it should really be default false, and then we can remove a lot of the with_wrapper: false in the factories

I believe the only place its true is in the model factory or the TextBlock, everywhere else is false?

app/services/cms/providers/strapi/factories/blocks_factory.rb
```diff
@@ -0,0 +1,210 @@
...
+              intro: ModelFactory.to_content_block(strapi_data[:intro], with_wrapper: false),
+              programme_activity_group_slug: strapi_data[:group][:data][:attributes][:slug]
+            )
+          end
+
+          def self.to_icon_block(strapi_data)
+            DynamicComponents::IconBlock.new(icons: strapi_data.map { to_icon(_1) })
+          end
+
+          def self.to_numeric_card_block(strapi_data)
```
msquance-stem, line 80: These next three methods are all `to_X_card_block`, just wondering if we should rename these as `to_X_card_array` or `to_X_cards` might be clearer these are returning an array rather than a block?

msquance-stem (changes requested):
Could of little points, the second one around naming is not so essential, but I think we can simplify the implementation of the to_content_block in a lot of places

app/services/cms/providers/strapi/factories/blocks_factory.rb
A-Wheeto: I'll rename these to to_X_card_array for clarity

PR #2269, opened 2025-01-27
Add support for GraphQL API for Starpi

msquance-stem:
## Status

* Review App link(s): 
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2935

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Created new client and connection for GraphQL
- Created new base client that handles data processing for both APIs (REST and GraphQL)
- Added conversion function in the factory to handle the different naming convention in of components and convert them to match the REST version
- New query system that holds the GraphQL for each component. There is an alias system to handle the fact that we have used the same name for attributes that have different definitions, prime example is title, which is often used by sometimes is optional. To deal with this you can add an alias on the front of the attribute and separate it with a double underscore (__), we just drop the first bit, so the latter just needs to be the name of the attribute


app/services/cms/providers/strapi/queries/components/blocks/horizontal_card.rb
```diff
@@ -0,0 +1,25 @@
...
+      module Queries
+        module Components
+          module Blocks
+            class HorizontalCard < BaseComponentQuery
+              def self.name = "ComponentBlocksHorizontalCard"
+
+              def self.base_fields
+                <<~GRAPHQL.freeze
+                  hozcard__title: title
+                  hozcar__textContent: textContent
```
A-Wheeto, line 13: Should these both be either hozcard or hozcar?

spec/services/cms/resource_spec.rb
```diff
@@ -15,14 +15,26 @@ def self.resource_key
...
+      allow(Rails.application.config).to receive(:strapi_connection_type).and_return(nil)
+      expect(described_class.send(:client)).to be_a Cms::Providers::Strapi::Client
+    end
+
+    it "default to rest client when not specified" do
+      allow(Rails.application.config).to receive(:strapi_connection_type).and_return("rest")
+      expect(described_class.send(:client)).to be_a Cms::Providers::Strapi::Client
+    end
+
+    it "default to rest client when not specified" do
```
A-Wheeto, line 29: Looks like the it statements on these tests need updating?

app/services/cms/providers/strapi/queries/base_query.rb
```diff
@@ -0,0 +1,124 @@
+module Cms
+  module Providers
+    module Strapi
+      module Queries
+        class BaseQuery
+          attr_accessor :resource_name, :resource_filter
+
+          QUERY_MAPS = {
```
A-Wheeto, line 8: It would be nice to put these into alphabetical order for better readability.

PR #2274, opened 2025-01-28
Fix links to primary course page on dashboard

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2940

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- New method in the user courses dashboard component to send the user to the correct course page depending on the programme
- Added additional tests in for the method

### DIFF
--- a/app/components/user_programme_course_bookings_with_asides_component.rb
+++ b/app/components/user_programme_course_bookings_with_asides_component.rb
@@ -20,7 +20,7 @@ class UserProgrammeCourseBookingsWithAsidesComponent < CmsWithAsidesComponent
 
   private
 
-  def button_link_text
+  def programme_link
     return primary_courses_path if @programme == Programme.primary_certificate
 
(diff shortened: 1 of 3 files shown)


app/components/user_programme_course_bookings_with_asides_component.rb
```diff
@@ -20,6 +20,12 @@ def before_render
 
   private
 
+  def button_link_text
```
msquance-stem, line 23: Could we just rename this to programme link, button_link_text made me think it was the text of the button, and programme_link is more generic, what you think?

A-Wheeto: Yep, on second look that's not very clear. Will change this.

msquance-stem (approved):
LGTM

PR #2278, opened 2025-02-03
Fixing description nil error on legacy activites

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2942

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Creating new safe method for activities to check for public copy first, then falling back to description is not included, and also allowing for nil values (to deal with legacy objects)


A-Wheeto (approved):
LGTM

PR #2294, opened 2025-02-12
Move primary computing glossary page to Strapi

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2956

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Added primary computing glossary table component into Strapi via graphql
- New CMS view component for the table
- New stubs, mocks and tests


app/services/cms/collections/primary_glossary_table.rb
```diff
@@ -0,0 +1,31 @@
+module Cms
+  module Collections
+    class PrimaryGlossaryTable < Resource
```
msquance-stem, line 3: I wonder if we should rename this to PrimaryGlossaryTableItems, just to make it clear its separate to the component? I wouldn't bother changing the name on Strapi, just the collection name here

msquance-stem (approved):
LGTM, just a naming change suggestion, and also a fix for lighthouse

app/services/cms/collections/primary_glossary_table.rb
A-Wheeto: Agreed, I will update this.

PR #2297, opened 2025-02-14
2962 fixing rack attack issues in rails 7 upgrade

msquance-stem:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2962

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Updating Cloudflare-rails
- It now has a simple method to check if the request was from cloudflare, so we have simplified the initialiser and the testing


A-Wheeto (approved):
LGTM

PR #2320, opened 2025-02-27
Remove mentions of subsidies

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2960

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Removed subsidies and funding copy from various pages around TC
- Updated mailers
- Updated tests


app/views/pages/home/_confidence.html.erb
```diff
@@ -12,14 +12,6 @@
       </p>
     </div>
 
-    <div class="ncce-confidence__card">
```
msquance-stem, line 15: We should probably flag this with comms, as looks a bit weird with one missing, so that they start coming up with a new one

spec/views/certificates/cs_accelerator/complete.html_spec.rb
```diff
@@ -67,11 +67,6 @@
       expect(rendered).to have_link("Find out more", href: "https://community.computingatschool.org.uk/events")
     end
 
-    it "has a subsidy section" do
-      expect(rendered).to have_link("this form", href: "https://static.teachcomputing.org/BACS+form+-+NCCE.docx")
```
msquance-stem, line 71: We should make a note to remove this from static in April (or check if they want it removing)

msquance-stem (approved):
LGTM, the comments are not about code, just things we should probably check with Comms.

PR #2328, opened 2025-03-04
Update routes to Strapi

msquance-stem:
## Status

* Review App link(s): 
* Closes: https://github.com/NCCE/teachcomputing.org-issues/issues/2993

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Updating routes to Strapi CMS
- Fixed some spacing issues with asides
- Fixing alignment on Icon row component
- Updated testing
- Removed obsolete hard coded pages


msquance-stem:
This should be almost good to go once @A-Wheeto has completed the enrolment resource card component.

One thing to double check before go live
- About has been swapped over to Strapi, but it doesn't appear to have had comms sign off according to the spreadsheet. So this should be checked

We have **NOT** migrated the careers support page. This is due to having to make a benchmark variants of all the components, and this seems excessive for one page.

PR #2332, opened 2025-03-07
Update Strapi page title with background colour and I belong flag

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2995

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Added options for background colour and I belong flag to page title component
- Updated page title mock
- Added graphql query test for page title


app/components/cms/page_title_component/page_title_component.html.erb
```diff
@@ -26,6 +26,10 @@
           <div class="page-title__media">
             <%= render Cms::ImageComponent.new(@title_image, show_caption: false) %>
           </div>
+        <% elsif @i_belong_flag %>
+          <div class="page-title__media i-belong-hero__area--logo">
```
msquance-stem, line 30: Should we change these classes to reflect the component, and then replicate the scss code into here, rather than relying on the global styles, as I think we should move away from them.

msquance-stem (changes requested):
Looks good, but I think we should bring the styles controlling the flag into the component, as it will be easier to find in the future, and then you don't have to worry about something external to the component changing it

app/components/cms/page_title_component.rb
```diff
@@ -1,11 +1,20 @@
...
     @title = title
     @sub_text = sub_text
     @title_image = title_image
     @title_video_url = title_video_url
     @status_message = status_message
+    @background_color = background_color
+    @i_belong_flag = i_belong_flag
+  end
+
+  def background_color_class
```
A-Wheeto, line 14: I realised we weren't setting a default background colour, existing pages using the component wouldn't have a background colour class. This method sets it to the default lime green.

msquance-stem: Could you just change the component definition to have "lime-green" as the default for the background color param? Saves you needing the method

msquance-stem (approved):
LGTM, one suggestion that would simplify your default colour

PR #2345, opened 2025-03-11
Removing type, date and location filter from searches

msquance-stem:
## Status

* Review App link(s): https://teachcomputing-pr-2345.herokuapp.com/
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2982

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Removed Type, Date and Location filters from the front end of the search system
- We have not removed the logic, this is because their might be a need to add these filters back in in the future, and we can leave the backend logic without issue.


A-Wheeto (approved):
LGTM

PR #2346, opened 2025-03-11
Strapi component - I Belong picture cards

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2997

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- New Strapi component for I Belong picture cards


app/services/cms/dynamic_components/card_wrapper.rb
```diff
@@ -1,19 +1,20 @@
...
 
-      def initialize(title:, intro_text:, cards_block:, cards_per_row:, background_color:, title_as_paragraph:)
+      def initialize(title:, intro_text:, cards_block:, cards_per_row:, background_color:, title_as_paragraph:, programme:)
         @title = title
         @intro_text = intro_text
         @cards_block = cards_block
         @cards_per_row = cards_per_row
         @background_color = background_color
         @title_as_paragraph = title_as_paragraph
+        @programme = Programme.find_by(slug: programme[:data][:attributes][:slug]) if programme
```
msquance-stem, line 13: You want to handle the `programme[:data][:attribute][:slug]` extraction in the factory, these data mappings should be entirely independent of the data models in Strapi

app/components/cms/card_wrapper_component.rb
```diff
@@ -1,13 +1,21 @@
 # frozen_string_literal: true
 
 class Cms::CardWrapperComponent < ViewComponent::Base
```
msquance-stem, line 3: Could we make a new card wrapper here, called ProgrammeCardWrapper, which could inherit this one, meaning we dont have to add programme to every card wrapper?

app/services/cms/providers/strapi/factories/blocks_factory.rb
```diff
@@ -147,14 +149,43 @@ def self.to_course_card_array(strapi_data)
...
+              DynamicComponents::ContentBlocks::ProgrammePictureCard.new(
+                title: card_data[:title],
+                image: to_image(card_data, :image, default_size: :medium),
+                text_content: to_content_block(card_data[:textContent]),
+                card_links: group_programme_card_links(card_data)
+              )
+            end
+          end
+
+          def self.group_programme_card_links(card_data)
```
msquance-stem, line 163: I like this, but I wonder if we could make a mapping object for, to make it more reusable?

So in app/services/cms we could have a MultiStateLink class, that takes in the these 6 params, and then logic for it is in that model, so we can reuse it if needed?

msquance-stem (commented):
Added a few ideas that could improve this

msquance-stem (approved):
LGTM

PR #2354, opened 2025-03-14
Changing CSA exam logic

msquance-stem:
## Status

* Review App link(s): https://teachcomputing-pr-2354.herokuapp.com/
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3012

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Due to the move to more online courses, we needed to change the logic of the CSA exam gateway, now you need 20 credits worth of face to face (this covers the legacy option where users could do two face to face course) or 30 credits, which covers the following scenarios
- 1 face to face & 1 online (10 + 20)
- 2 online (20 + 20)
- 1 residential (40)

Dashboard has been update to reflect the single poll of options for courses, rather than required and recommended
Removed a load of old logic related to testing
While was doing this cleared the bullet warnings off dashboards relating to eager loading warnings

### DIFF
--- a/spec/view_objects/csa_dash_spec.rb
+++ b/spec/view_objects/csa_dash_spec.rb
@@ -46,7 +46,7 @@ RSpec.describe CSADash do
     end
   end
 
-  describe "#has_enough_activities_for_test" do
+  describe "#enough_credits_for_test?" do
     it "calls method on programme" do
       allow(csa).to receive(:enough_credits_for_test?)
       dash.has_enough_activities_for_test
--- a/spec/view_objects/csa_dash_spec.rb
+++ b/spec/view_objects/csa_dash_spec.rb
@@ -13,36 +13,11 @@ RSpec.describe CSADash do
     end
   end
 
-  describe "#compulsory_achievement" do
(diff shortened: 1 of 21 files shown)


spec/view_objects/csa_dash_spec.rb
```diff
@@ -71,25 +46,11 @@
...
-      allow(user).to receive(:programme_pathway) { pathway }
-
-      dash.supplementary_activities_for_user
-      expect(pathway)
-        .to have_received(:supplementary_activities_for_user)
-        .with(user)
-    end
-  end
-
   describe "#has_enough_activities_for_test" do
```
A-Wheeto, line 49: Does this need updating to the correct method name "enough_credits_for_test?"

A-Wheeto (approved):
Just one describe block in a test that needs updating, the rest all looks good though!

spec/view_objects/csa_dash_spec.rb
msquance-stem: Good spot!

PR #2374, opened 2025-03-24
Adding replaced_by logic to course cards

msquance-stem:
## Status

* Current Status: Ready for  tech review
* Review App link(s): https://teachcomputing-pr-2374.herokuapp.com/
* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3000

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

- Implemented the replaced by check on the data model for course cards
- This will replace the link and the icon information, but leave the course title and description the same
- This is intended as a temporary protection in case a course card is missed, but should be updated in Strapi to correct course instance, hence sentry notification


A-Wheeto (approved):
LGTM

PR #2386, opened 2025-03-28
Strapi component - Accordion

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3001

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- New accordion Strapi component


app/components/cms/accordion_section_component/accordion_section_component.scss
```diff
@@ -0,0 +1,16 @@
+.cms-accordion-section-component {
+  .govuk-accordion {
+    font-family: $font-family-body;
+  }
+
+  .govuk-accordion__section-button {
+    font-weight: 500;
```
msquance-stem, line 7: You might want to look a the govuk-font mixins to do this, rather than setting the font-size and weight directly

https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-font

msquance-stem (commented):
Looks good to me, I think my main point is that the blocks should be plural throughout, as its not clear its an array, I put a similar comment on the Strapi side.

I added a link to the typography helper methods that you should look at rather than setting the font size.

This also needs rebasing to match the updated structure.

app/components/cms/accordion_block_component.rb
```diff
@@ -0,0 +1,29 @@
...
+  erb_template <<~ERB
+    <div class="govuk-accordion__section">
+      <div class="govuk-accordion__section-header">
+        <h2 class="govuk-accordion__section-heading">
+          <span class="govuk-accordion__section-button">
+            <%= @heading %>
+          </span>
+        </h2>
+        <% if @summary_text %>
+          <div class="govuk-accordion__section-summary govuk-body-s" id="accordion-with-summary-sections-summary">
```
msquance-stem, line 13: This id and the one below might end up being replicated as there are multiple blocks in the section, can we add something make them unique?

msquance-stem (commented):
Just a couple of things that need looking at but otherwise looks good

app/services/cms/providers/strapi/factories/content_block_factory.rb
```diff
@@ -34,6 +36,14 @@ def self.to_link(strapi_data)
             )
           end
 
+          def self.to_accordion_block(strapi_data)
+            Models::DynamicComponents::ContentBlocks::AccordionBlock.new(
```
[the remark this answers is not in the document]
A-Wheeto: Good spot, this won't be needed and I'll remove.

msquance-stem (approved):
LGTM

PR #2397, opened 2025-04-03
Fixing flaky tests

msquance-stem:
## Status

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?

Most were caused by using specific stem activity codes, removed the need for the specific codes in all but one place, and in that one place but a handle to retrieve if already exists


A-Wheeto (approved):
LGTM

PR #2415, opened 2025-04-29
Add enrol button to resource cards

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3038

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- This was going to be a simple "add button to existing component", but as the resource cards are a content block in Strapi I was unable to add a nested component. I also discovered the text content on the cards change depending on the user state - so a new component Programme Resource Cards has been created
- Also refactored the enrol button into the buttons namespace


app/services/cms/providers/strapi/factories/blocks_factory.rb
```diff
@@ -318,16 +320,46 @@ def self.to_programme_picture_card_array(strapi_data, programme)
...
+            end
+
             Models::DynamicComponents::Blocks::ProgrammeCardWrapper.new(
               title: strapi_data[:sectionTitle],
               intro_text: to_content_block(strapi_data[:introText].presence || []),
-              cards_block: to_programme_picture_card_array(strapi_data[:programmeCards], programme),
+              cards_block: cards_block,
               cards_per_row: strapi_data[:cardsPerRow].presence || 3,
-              background_color: extract_color_name(strapi_data, :bkColor),
+              background_color: extract_color_name(strapi_data, :bkC),
```
A-Wheeto, line 362: I know this is a weird name for background color - but I was limited to 3 characters due to Strapi's character length

app/components/cms/programme_resource_card_component.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+class Cms::ProgrammeResourceCardComponent < ViewComponent::Base
```
msquance-stem, line 3: Something that you might not want to do in this ticket, but as a future tech debt ticket, is to align this with the `ResourceCardComponent`, as there is a lot of repetition between the two. Then this can mostly just be overrides of inherited behaviour.

If you were to have a method in the ResourceCardComponent `text_content` that just returned `@body_text` and then this inherited that class, you can just override that method. Then all this would need to do is call `render_parent`. You would need to swap the ResourceCardComponent to use the NcceButton model, as then in both the button will have a render method. That could be done in Strapi, or just in the factory in rails to wrap the `button_text` and `button_link` in that class. It would also give you in the icon in this component, as you know they will want that eventually.

app/components/cms/programme_resource_card_component/programme_resource_card_component.scss
```diff
@@ -0,0 +1,6 @@
+.cms-programme-resource-card-component {
+  background-color: $white;
```
msquance-stem, line 2: You are using white-bg class and setting the background colour here, I would stick to one

app/services/cms/providers/strapi/factories/base_factory.rb
```diff
@@ -77,6 +77,14 @@ def to_ncce_button(strapi_data)
               )
             end
 
+            def to_enrol_button(strapi_data)
+              Models::DynamicComponents::Buttons::EnrolButton.new(
+                logged_out_button_text: strapi_data[:loggedOutButtonText],
+                logged_in_button_text: strapi_data[:loggedInButtonText],
+                programme_slug: strapi_data[:programme][:data][:attributes][:slug]
```
msquance-stem, line 84: Do we need some catch to handle them not providing a programme slug, as we cannot make it required in Strapi? Should we let it error, or just not render the button?

app/services/cms/providers/strapi/factories/blocks_factory.rb
msquance-stem: I added a comment in Strapi about a solution that might help with this. I would also suggest given that this method handles the programme picture card section that its worth adding support for background colour to that

app/services/cms/providers/strapi/factories/blocks_factory.rb
```diff
@@ -318,16 +320,46 @@ def self.to_programme_picture_card_array(strapi_data, programme)
             end
           end
 
+          def self.to_programme_resource_card_array(strapi_data, programme_slug)
+            programme = Programme.find_by(slug: programme_slug)
+
+            strapi_data.map do |card_data|
+              button = if card_data[:loggedOutButtonText] && card_data[:loggedInButtonText]
```
msquance-stem, line 327: Could you move the check for fields into the `to_enrol_button` method, to reduce the complexity of this method?

app/services/cms/providers/strapi/queries/aside.rb
```diff
@@ -4,8 +4,8 @@ module Strapi
       module Queries
         class Aside
           CONTENT_COMPONENTS = [
-            Components::ContentBlocks::EnrolButton,
             Components::ContentBlocks::FileLink,
```
msquance-stem, line 7: FileLink needs moving down this list to keep the ordering

spec/components/cms/programme_resource_card_component_spec.rb
```diff
@@ -0,0 +1,174 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Cms::ProgrammeResourceCardComponent, type: :component do
+  let(:user) { create(:user) }
+  let!(:programme) { create(:i_belong) }
+
+  context "when no logged in user" do
+    let(:logged_out_text_content) { Cms::Mocks::Text::RichBlocks.as_model }
```
msquance-stem, line 10: You could move these three lets to the top of the test, as I believe they are in every test?

msquance-stem (commented):
Generally looking pretty good, I think there is a few things that could just tidy up the implementation a bit.

app/components/cms/programme_resource_card_component.rb
```diff
@@ -0,0 +1,35 @@
...
+    @button = button
+  end
+
+  def wrapper_classes
+    classes = ["cms-programme-resource-card-component", "white-bg"]
+    classes << cms_color_theme_class(@color_theme, "top") if @color_theme
+    classes
+  end
+
+  def text_content
```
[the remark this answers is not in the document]
A-Wheeto: Good spot, these should all be required fields in Strapi and I'll update this.

app/components/cms/programme_resource_card_component.rb
A-Wheeto: Great idea, I'll put in a ticket to look at this at a later date.

msquance-stem (approved):
LGTM

PR #2420, opened 2025-05-08
Updating time_diff_words method to handle months that aren't 30 days

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3048

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- The time_diff_words method was failing tests as it assumed that months would have 30 days, but as we are now comparing back to February which only had 28 days the method needed updating to account for this.

### DIFF
--- a/app/services/cms/models/collections/email_template.rb
+++ b/app/services/cms/models/collections/email_template.rb
@@ -49,10 +49,9 @@ module Cms
           achievements = user.sorted_completed_cpd_achievements_by(programme: @programme)
           if achievements.any?
             achievement = achievements.last
-            last_cpd_completed_ago = distance_of_time_in_words(achievement.updated_at, DateTime.now)
 
             merges += [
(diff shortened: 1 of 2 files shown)


markjs:
Have you considered using Rails's built in `distance_of_time_in_words` method for doing exactly this? Even if you don't want to use exactly what it outputs for every case there may be a tidier way to use this method along with a couple of conditionals that is hopefully more robust. https://apidock.com/rails/ActionView/Helpers/DateHelper/distance_of_time_in_words

markjs:
@A-Wheeto Looks good. I pointed you to `distance_of_time_in_words` although since it's based on the current time you could swap that out for `time_ago_in_words` to save a tiny bit of complexity but that's not super necessary 😄

markjs:
https://apidock.com/rails/ActionView/Helpers/DateHelper/time_ago_in_words

It just calls the same method behind the scenes but with a fixed `to_time`

PR #2425, opened 2025-05-16
Move footer to Strapi

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3039

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Replaced existing footer with CMS Footer component


app/components/cms/footer_component.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class Cms::FooterComponent < ViewComponent::Base
+  attr_reader :data
+
+  def initialize(data:)
+    @data = data
+  end
+
+  def company_logo
+    data_models_by_type(Cms::Models::Images::Image).first
```
markjs, line 11: Seems a bit risky fetching these values based on their order. It looks like there's a key being used elsewhere, perhaps this could be refactored to make use of the key here too instead of first/second?

app/components/cms/footer_component.rb
```diff
@@ -0,0 +1,39 @@
...
+    data_models_by_type(Cms::Models::Meta::FooterLinkBlock).first&.process_blocks
+  end
+
+  private
+
+  def data_models_by_type(type)
+    data.select { |model| model.is_a?(type) }
+  end
+
+  def find_text_field(index)
```
markjs, line 36: This method seems a little unnecessary. The `company_url` and `funding_url` methods could be rewritten to be more similar to the `company_logo` and `funding_logo` methods to just have the `data_models_by_type` call directly in them and this extracted method wouldn't be needed.

markjs (commented):
A couple of comments on some of the lookups in the component, otherwise looks good 👍

A-Wheeto:
Thanks @markjs - on second look you have a good point on the `get_model` calls in the view. I'll tidy this up.

PR #2431, opened 2025-06-09
Remove template caching error from Sentry

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3056

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Updated template caching to log info into rails logger rather than Sentry error
- Added Sentry error if templates aren't retrieved from Dynamics


matoakley (approved):
LGTM!

PR #2433, opened 2025-06-12
Adding additional testing for exception handling

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3062

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Added additional testing for exception handling in the fetch from achiever job
 
### DIFF
--- a/spec/jobs/achiever/fetch_users_completed_courses_from_achiever_job_spec.rb
+++ b/spec/jobs/achiever/fetch_users_completed_courses_from_achiever_job_spec.rb
@@ -5,6 +5,9 @@ RSpec.describe Achiever::FetchUsersCompletedCoursesFromAchieverJob, type: :job d
 
   let(:user) { create(:user) }
   let(:activity_one) { create(:activity, stem_course_template_no: "92f4f86e-0237-4ecc-a905-2f6c62d6b5ae") }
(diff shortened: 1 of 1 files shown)


matoakley (approved):
Nice work @A-Wheeto!

One thing to consider, although not a blocker at all, is that we're tying ourselves in to Sentry as an exception tracker by adding explicit calls to the Sentry gem throughout the job.

If we decided to migrate to another exception tracker in the future (e.g. Honeybadger) then we'd need to find and replace each instance of the call to the Sentry gem.

This problem is often solved by wrapping the calls to the external gem in a local wrapper, e.g. in the this case you could create an `ExceptionTracker` class with a `capture_exception` function which then calls Sentry.

With a local abstraction in place, we'd only need to update the `ExceptionTracker` class if we ever moved to a new provider or if Sentry made a breaking change to their gem's API.

This pattern is typically referred to as Dependency Inversion Principle.

Not necessarily suggesting that we need to implement this now, just a thought for the future.

A-Wheeto:
Thanks @matoakley  — really good point, and definitely something worth considering. I'll add a tech debt ticket to the backlog so we can revisit this down the line. Since there’s a chance the website could be handed off to another owner in the future (who might use a different exception tracker), it makes sense to maintain that layer of abstraction.

PR #2436, opened 2025-06-16
Administrate improvement and updates

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/2739

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
- Removed hubs and hub regions from administrate as no longer used
- Organised routes so dashboards now show in alphabetical order
- Added a new programme dashboard that shows an overview of
  - programme activity groupings
  - pathways
  - assessments
- Minor updates to reports dashboard

image

### DIFF
--- a/spec/requests/admin/programmes_spec.rb
+++ b/spec/requests/admin/programmes_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe "Admin::ProgrammesController" do
       end
 
       it "should render programme activity groupings activity count" do
-        expect(response.body).to include(programme.programme_activity_groupings.first.activities.count.to_s)
+        expect(response.body).to include(programme.programme_activity_groupings.first.activities.size.to_s)
       end
 
       it "should render activities" do
@@ -89,7 +89,7 @@ RSpec.describe "Admin::ProgrammesController" do
       end
 
       it "should render pathway activity count" do
-        expect(response.body).to include(programme.pathways.first.pathway_activities.count.to_s)
+        expect(response.body).to include(programme.pathways.first.pathway_activities.size.to_s)
       end
 
       it "should render activities" do


app/views/admin/programmes/show.html.erb
```diff
@@ -0,0 +1,76 @@
...
+    <table>
+      <tr>
+        <th>Title</th>
+        <th>Activity count</th>
+        <th>Activities</th>
+      </tr>
+      <% @programme.pathways.each do |pathway| %>
+        <tr>
+          <td><%= pathway.title %></td>
+          <td><%= pathway.pathway_activities.count %></td>
```
barrywoolgar, line 46: Using `.count` on an association you've eager loaded will be triggering an extra SQL query unnecessarily. In Rails it's best to prefer `.size` over `.count` as the former will first check if the collection has already been loaded into memory, and count that instead.

app/controllers/admin/programmes_controller.rb
```diff
@@ -0,0 +1,20 @@
+module Admin
+  class ProgrammesController < Admin::ApplicationController
+    def index
+      @programmes = [
+        Programme.cs_accelerator,
+        Programme.primary_certificate,
+        Programme.secondary_certificate,
+        Programme.i_belong,
+        Programme.a_level
+      ].compact.sort_by(&:title)
```
barrywoolgar, line 10: This raised an eyebrow at first, but I then realised this is just 5 records being sorted in memory, not 5 sets of records.

barrywoolgar (approved):
One micro-optimisation of note, but otherwise this looks good to me. Thanks Adam!

app/views/admin/programmes/show.html.erb
A-Wheeto: Good tip, I'll update this to `.size`, thanks Barry!

PR #2456, opened 2025-08-27
Update controllers to use params

A-Wheeto:
## Status

* Closes https://github.com/NCCE/teachcomputing.org-issues/issues/3074

## Review progress:

- [ ] Browser tested
- [ ] Tech review completed

## What's changed?
The following controllers were using the `request` object in the response:
- class_marker/webhooks_controller.rb
- dynamics/webhooks_controller.rb
- concerns/curriculum/rateable.rb

This is outdated and we should be using `params`.

As part of this tech debt, I've added some additional checking into the classmarker controller to ensure the required data is present before proceeding to the job.


markjs (approved):
Looks good, certainly much more robust than before and should give a more helpful error in the case of missing webhook params. Nice one 👍

PR #2543, opened 2026-05-13
Teach computing to Strapi 5

A-Wheeto:
- base_client.rb — dual-format response parsing (GraphQL _connection + REST), documentId || id fallback
 - graphql_client.rb — removed inner `[:data]` unwrap, improved nil check
 - base_query.rb — removed `data.attributes` wrappers from both queries
 - resource.rb — `all_records` break condition fixed
 - All factory files — relation access flattened from `[:data][:attributes][:field]` to `[:field]`
 - shared_fields.rb — `image_fields` slimmed to 4 fields, `file_fields` added
 - file_link.rb query — switched from `image_fields` to `file_fields`
 - enrichment_list.rb — flattened query structure
 - All 20+ mock files — `data.attributes` nesting removed


.gitignore
```diff
@@ -51,6 +51,7 @@ public/sitemap.xml.gz
 .history
 *.code-workspace
 .irb_history
+.claude
```
matoakley, line 54: I don't think it's common practice to add the entire `.claude` directory to `.gitignore` as there are certain settings and/or skills that we may wish to share in the future. But not a big problem for now.

matoakley (approved):

PR #2549, opened 2026-05-29
Handle expired mylearning courses on teach computing

A-Wheeto:
ENG-1646


app/views/courses/_aside-booking.html.erb
```diff
@@ -140,7 +153,7 @@
 
   <%= render ProviderLogosComponent.new(dashboard: false, no_top_border: current_user.blank? || @course.online_cpd)%>
 
-  <% if current_user && user_achievement_state(current_user, @activity) == :enrolled && @course.online_cpd %>
+  <% if current_user && user_achievement_state(current_user, @activity) == :enrolled && @course.online_cpd && !always_on_access_expired?(current_user, @activity) %>
```
matoakley: Would it be worth adding a method to `app/helpers/courses_helper.rb` called `should_rejoin_course?` and moving the logic there to tidy up the template?

A-Wheeto: Yep, good shout. This line is a bit messy.

matoakley (approved):

PR #2559, opened 2026-06-11
Upgrade to Rails 8.1

A-Wheeto:
Upgrades Rails from 7.1 to 8.1.3 and resolves all breaking changes introduced by the upgrade.

**Gem changes**
- rails bumped from ~> 7.1.0 to ~> 8.0 (resolved: 8.1.3)
- view_component upgraded from 3.x to 4.12.0 (required by Rails 8; 3.x had a < 8.0 constraint)
- administrate unpinned from ~> 0.20.0 to current (1.0.0; now supports Rails 8)
- cloudflare-rails upgraded to 7.0.0 (6.x had a < 8.1.0 constraint)
- administrate-field-nested_has_many removed (incompatible with Administrate 1.0; its gemspec requires administrate < 1)
- rspec-rails bumped to 8.0.4
- ostruct added explicitly (extracted from stdlib in Ruby 3.5+)
- sprockets-rails and sassc-rails added as explicit app dependencies - Administrate 1.0 dropped its own transitive dependency on these, so the app now owns them directly for its Sprockets-based admin dashboard styles
- connection_pool pinned to ~> 2.4 - Sidekiq 7.3.9 calls `ConnectionPool::TimedStack#pop` with a positional argument; connection_pool 3.0 made this keyword-only, which crashes Sidekiq's scheduler thread on startup. Sidekiq's own fix for this only lands at 8.1.0 (a major bump we're not taking here), so this pin is required for as long as we stay on Sidekiq 7.x

**Breaking changes fixed**
- Enum syntax (app/models/badge.rb) - Rails 8 removed the old hash-rocket enum form. Updated to positional syntax
- Time arithmetic (_csa-test.html.erb, _exam.html.erb) - can_take_test_at returns an integer (seconds). Rails 8.1 removed implicit Integer-to-Duration coercion when adding to a Time/TimeWithZone, so .seconds is now explicit
- Routes (config/routes.rb) - Rails 8 added validation rejecting non-standard action names in `resources ... only:`. perform_sync/perform_reset were incorrectly listed there; moved to a `member do` block with explicit named routes instead
- ViewComponent 4 - super → super() in ModalComponent and RelatedLinksComponent (VC4 removed the catch-all initializer that used to make bare `super` safe); preview config keys renamed (previews.paths, previews.route, previews.enabled); image_pack_tag delegation added to initializer for Webpacker compatibility until Phase 3
- ViewComponent generator namespace (lib/generators/strapi/component_generator.rb). ViewComponent 4 renamed the generator from "component" to "view_component:component". Rails' Generators.invoke now calls exit 1 on a missing generator (previously just warned), which was raising SystemExit in RSpec before-hooks and killing the test suite mid-run with variable test counts

**Notes**
- config.load_defaults remains at 7.1 - adopting Rails 8.0 defaults is deferred to a follow-on phase (same strategy used for the 7.1 defaults in the preceding branch)
- config.sass.style/line_comments were briefly dropped from production/staging/test during an environment-config cleanup pass and have been restored - these are unrelated to Rails 8 (they're sassc-rails's own config, independent of Rails version) and control minification/debug-comments for the Sprockets-compiled Administrate CSS specifically, not the main site's Webpacker CSS
- Phase 3 (replacing Webpacker with importmap + dartsass) will be completed in ENG-1744


config/routes.rb
```diff
@@ -32,7 +32,7 @@
     end
     resources :sent_emails, only: %i[index show]
     resources :support_audits, only: %i[index show update edit]
-    resources :users, only: %i[index create show edit perform_sync perform_reset update] do
+    resources :users, only: %i[index create show edit update] do
```
matoakley, line 35: Are we certain that `/perform_sync` and `/perform_reset` are never called without the `user_id` param?

A-Wheeto: Thanks, Matt. That's a really good spot. Although these actions are performed from within administrate and would have no front end effect, they need addressing. I'll put a fix together.

app/controllers/admin/users_controller.rb
```diff
@@ -10,7 +10,7 @@ def perform_sync
       )
     end
 
-    def perform_reset
+    def perform_reset_tests
```
matoakley, line 13: What was the reason for renaming this method?

A-Wheeto: The original code had a mismatch. The route was named `perform_reset` but the controller action was `perform_reset_tests`. I made it more consistent, but then decided to rename back to `perform_reset_tests` because that's what it actually does (calls reset_tests), and now both the route and action are in line.

PR #2580, opened 2026-07-23
Course component shows incorrect enrolled and completed dates

A-Wheeto:
ENG-1648


bompstable (approved):
Looks good to me. This does assume that there will never be another update to the achievement though - if that happened, the completion date would presumably be wrong.

A-Wheeto:
Thanks, Max. I did consider that and there's nothing in the code that would transition a course to a different state after it reaches complete, nor would we want to.

PR #2587, opened 2026-07-27
Create new secondary standalone activities

A-Wheeto:
- CQF and the "implement and evaluate your professional development" activity are extracted from their existing activity groupings and shown as standalone dashboard sections on the secondary certificate page, above the professional development groups list
- Each is modelled as a single-activity `ProgrammeActivityGrouping` (new `CqfAssessment`/`ProfessionalDevelopmentActivity `STI subclasses) rather than a new component - reuses
  the existing community activity list/component
- Both activities are duplicated with a `-secondary` slug so completion tracking is independent per programme (previously shared with the Primary certificate)

I've also included a UI tweak to the show page:
- Collapse the professional development component on page load - previously, the component loaded expanded. With the 2 new activity components introduced in this ticket, they rendered far down the page. Loading collapsed allows the users to quickly see the objectives easier without having to scroll too far down the page

### DIFF
--- a/app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
+++ b/app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
@@ -1,0 +1,11 @@
+<% title_classes = ["govuk-body", "ncce-activity-list__title"] %>
+<% title_classes << "govuk-!-margin-top-8" if local_assigns[:margin_top] %>
+<% title_classes << add_group_complete_icon_class(current_user, group) %>
+
+<p id="<%= group.id %>" class="<%= title_classes.select(&:present?).join(" ") %>">
+  <strong><%= group.title %></strong>
+</p>
+<%= render CommunityActivityListComponent.new(
+  programme_activity_grouping: group,
+  community_achievements: @community_achievements
+) %>
--- a/app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
+++ b/app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
@@ -1,5 +1,4 @@
 <% title_classes = ["govuk-body", "ncce-activity-list__title"] %>
-<% title_classes << "govuk-!-margin-top-8" if local_assigns[:margin_top] %>
 <% title_classes << add_group_complete_icon_class(current_user, group) %>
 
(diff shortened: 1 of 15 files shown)


app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
```diff
@@ -0,0 +1,10 @@
+<% title_classes = ["govuk-body", "ncce-activity-list__title"] %>
+<% title_classes << add_group_complete_icon_class(current_user, group) %>
+
+<p id="<%= group.id %>" class="<%= title_classes.select(&:present?).join(" ") %>">
```
matoakley, line 4: Assuming that IDs are integers then this will set `id="1"` which, given that IDs are expected to be unique per page, is a bit risky. Especially given that this is a partial which will be included within another template.

Typically you'd namespace the ID with the resource name - e.g. `id="group-1"` which is less likely to clash with other IDs on the page.

matoakley (changes requested):
One comment but otherwise looks great.

app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
A-Wheeto: Thanks, Matt. Good one to check. `group.id` is the UUID of the group here, rather than an integer, so we'll never get a clash. 

This also matches the same pattern already used elsewhere on this page, so it keeps things consistent.

matoakley: > so we'll never get a clash

That's not strictly true but agree it's very low risk.

The other risk is that the same record appears elsewhere on a page, so maybe it'd have been safer to namespace it by the partial name "standalone-activity-group-1" but if you are confident that the same record will never appear twice then I'm happy to approve.

matoakley (approved):

app/views/certificates/secondary_certificate/_standalone_activity_group.html.erb
A-Wheeto: Very good point. Currently the CQF/PD groups are explicitly excluded from the professional development groups list, so we'd never render the same group twice on the page. I'd be happy to call this safe. I'll certainly keep this in mind if there's any future changes.

PR #2601, opened 2026-08-06
Fix 500 error when searching achievements in admin panel

A-Wheeto:
Administrate's default search builds SQL directly against a dashboard's `ATTRIBUTE_TYPES` fields marked searchable. `stem_activity_code` is a `Field::String` but isn't a real column on achievements (it's delegated to the associated Activity), so any search raised `PG::UndefinedColumn`

- Changed `stem_activity_code` to `Field::String.with_options(searchable: false)` so it still displays but is excluded from the search query
- Added request spec coverage for searching by activity code, partial activity title, and user email, since the original 500 was only caught manually


bompstable:
I asked Claude to review this PR branch and it came up with the following. These are definitely in decreasing order of importance but 1 looks like we should act on it.

1. NonSearchableStringField is unnecessary — the codebase already has an idiom for this. [assessment_attempt_dashboard.rb:10](vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/app/dashboards/assessment_attempt_dashboard.rb#L10) solves the identical problem (current_state is a statesman delegate, not a column) with Field::String.with_options(searchable: false), which Field::Deferred#searchable? honours. The new class also has two costs: html_class becomes non-searchable-string-field, so the rendered classes change from cell-label--string/attribute-data--string to the new name (silent styling change on the index/show pages); and the local_partial_prefixes override is redundant — partial_prefixes already walks the superclass chain and find_partial_prefix picks the first existing template, so fields/string would be found anyway. Suggest deleting [app/fields/non_searchable_string_field.rb](vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/app/fields/non_searchable_string_field.rb) and using Field::String.with_options(searchable: false).

2. No test covers the regression. [spec/requests/admin/achievements_spec.rb:28](vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/spec/requests/admin/achievements_spec.rb#L28) only does get admin_achievements_path with no search param, which is exactly why the 500 got through. Worth adding cases for get admin_achievements_path(search: activity.stem_activity_code), a partial title, and user.email, asserting the right record comes back — cheap, and it pins the join behaviour too.

3. Nit: GroupedActivityListField.permitted_attribute ([grouped_activity_list_field.rb:14-16](vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/app/fields/grouped_activity_list_field.rb#L14-L16)) is now dead weight — BelongsTo.permitted_attribute derives activity_id from the reflection, and BaseDashboard#permitted_attributes passes the resource_class: it needs. Can be deleted.

4. Optional: admins tend to search by TRN — teacher_reference_number and stem_user_id are both columns on users and could join the user searchable_fields.

Two things worth knowing but not blocking:

Unintended win: BelongsTo.eager_load? is true, so activity now lands in collection_includes — the index page previously N+1'd one activities query per row for the stem_activity_code and activity columns. Nice side effect, worth a line in the PR description.
Search cost: the query is now LOWER(CAST(col AS CHAR(256))) LIKE '%term%' across five columns over left joins to users and activities — unindexable, so it'll scan. Fine at admin scale, just not free as achievements grows.

A-Wheeto:
Thanks, Max. I think Claude's assessment is fair here and `NonSearchableStringField` was an unnecessary addition.

I've updated to use `Field::String.with_options(searchable: false)` and also added some additional testing.

bompstable:
Thanks, Adam. That looks much better and answers another question I would have asked about the original fix.<br><br>However, Claude now says the following:

One real problem with the new specs:

**The email test can't fail.** Administrate's search bar echoes the term into the page — `value="<%= search_term %>"` in the gem's `application/_search.html.erb`. So at [achievements_spec.rb:66](<vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/spec/requests/admin/achievements_spec.rb#L66>), `expect(response.body).to include(matching_user.email)` is satisfied by the search input itself, and would still pass if the query returned zero rows. The other two tests dodge this by chance: `"CP123"` and `"Matching activ"` are echoed, but neither is the string being asserted. Fix by asserting on something the term isn't a substring of — the row's activity title:

```
it "finds achievements by user email" do
  get admin_achievements_path(search: matching_user.email)

  expect(response).to have_http_status(:ok)
  expect(response.body).to include("Matching activity")
  expect(response.body).not_to include("Other activity")
end
```

That keeps the `not_to include("other-user@example.com")` line meaningful too if you want to keep it — negative assertions aren't affected by the echo.

Two nits, take or leave:

* The outer `before { get admin_achievements_path }` at [achievements_spec.rb:28](<vscode-webview://1v2lr3pcosgcpu8gmsg7k53vjoa23bth286t6b6kfpkuq7725pf9/spec/requests/admin/achievements_spec.rb#L28>) still runs for each search example — and *before* the inner `before` creates the records — so every search test fires a wasted request against an empty table first. A sibling `describe "GET #index with a search term"` would avoid inheriting it.
* `"CP123"` is a `LIKE` substring of factory-generated codes `CP1230`–`CP1239` (the sequence starts at 1000 and the counter is process-global), so a stray row can join the results. It can't fail the current assertions — stray rows get titles like `activity142` — but a code with no numeric neighbours would be sturdier.

Worth noting these are the first examples in this file to render index rows with data at all (the pre-existing `GET #index` test never referenced the `achievement` let), so they cover more than just search.

A-Wheeto:
Good spot. I've updated this test. Thanks, Max.

